PR #26, opened 2022-09-27
chore: cleanup directories

ryanwolhuter:
### Summary

This PR looks big but it changes no logic. It's only purpose is to do some much-needed organizing of files and folders.

I've added index files for all the directories to make imports cleaner. Sticking to this pattern will be easier going forward now that the groundwork has been laid.

* Create indexes for all directories and update imports


md0x (approved):
LGTM
Imports are much cleaner!

PR #27, opened 2022-09-27
bug: storybook browser polyfills

ryanwolhuter:
### Summary

Storybook doesn't get access to all the nodejs browser polyfills that are available to the main app because of the way it handles imports (long story). Instead we need to manually add them in the storybook webpack config.

I've opted to add all of the obvious ones in advance, since they don't add size to the bundle.


md0x (approved):
LGTM

PR #28, opened 2022-09-28
bug: numeric input behavior

ryanwolhuter:
### Summary

Some bugs were identified with numeric inputs in the dapp. 

[Fields error on too many decimals](https://linear.app/uma/issue/UMA-474/staking-field-errors-on-too-many-decimals)

- Add a helper hook that lets us specify a maximum number of decimals for an input (default 18)

[Scroll buttons allow negative values](https://linear.app/uma/issue/UMA-473/staking-scroll-buttons-allows-setting-negative-value)

These scroll buttons are not meant to be there at all (they appear only on FireFox)

- Hide scroll buttons with CSS in component
- Add an `allowNegative` flag to inputs for cases where negatives are not allowed

[Max staking button is confusing](https://linear.app/uma/issue/UMA-472/max-button-in-staking-not-functional)

- After re-thinking this I realized that my original logic that used the `tokenAllowance` as the max value was a poor decision. It's much easier and clearer to the user to just always use the `unstakedBalance` as the max.


components/GlobalStyle/GlobalStyle.ts
```diff
@@ -164,6 +164,9 @@ a:not([class]) {
 :where(input[type="file"]) {
   cursor: auto;
 }
+:where(input[type="number"]) {
```
ryanwolhuter, line 167: Hides the "scroll buttons" on firefox

components/Input/AmountInput.tsx
```diff
@@ -1,24 +1,36 @@
...
+  const onChange = useOnChange(onInput, maxDecimals, allowNegative);
+
   return (
     <_Wrapper aria-disabled={disabled}>
       <_Input
         value={value}
         onChange={onChange}
         disabled={disabled}
-        type="number"
+        type="text"
```
ryanwolhuter, line 33: There are several annoying things about using a `number` type input, the main one here being that it will automatically replace "." with "," as the decimal separator if the locale is one that uses "," for decimals. This inconsistent behavior is ugly and also breaks things because we use "," to separate thousands as well.

components/VotesTable/VotesTableRow.tsx
```diff
@@ -29,6 +29,7 @@ export function VotesTableRow({
   const { signer } = useWalletContext();
 
   const { decodedIdentifier, title, origin, options, isCommitted, isRevealed, decryptedVote, correctVote } = vote;
+  const maxDecimals = getPrecisionForIdentifier(decodedIdentifier);
```
ryanwolhuter, line 32: Use the decimal precision for the identifier as the max decimals (this can be different than 18 for some)

hooks/helpers/useOnChange.ts
```diff
@@ -0,0 +1,42 @@
+import { useErrorContext } from "hooks";
+import { ChangeEvent } from "react";
+
+export function useOnChange(
+  onInput: (value: string, ...args: unknown[]) => void,
+  maxDecimals: number,
+  allowNegative: boolean,
+  type = "number"
+) {
+  const { addErrorMessage, removeErrorMessage } = useErrorContext();
```
ryanwolhuter, line 10: handles showing and hiding error for too many decimals

hooks/helpers/useOnChange.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  return (event: ChangeEvent<HTMLInputElement>) => {
```
ryanwolhuter, line 12: returns a change handler function which takes the `event`

hooks/helpers/useOnChange.ts
```diff
@@ -0,0 +1,42 @@
...
+    const value = event.target.value;
+
+    if (type !== "number") {
```
ryanwolhuter, line 15: For non-numeric we just update the value as normal

md0x (approved):
LGTM

hooks/helpers/useOnChange.ts
```diff
@@ -0,0 +1,42 @@
+import { useErrorContext } from "hooks";
+import { ChangeEvent } from "react";
+
+export function useOnChange(
```
daywiss, line 4: this is really genericly named, but it seems like it does something really specific, might want to name it something more relevant

hooks/helpers/useOnChange.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (hasDecimals) {
+      const decimals = value.split(".")[1];
+      const hasTooManyDecimals = decimals.length > maxDecimals;
+      if (hasTooManyDecimals) {
+        addErrorMessage(decimalsErrorMessage);
+        return;
+      }
+    }
+
+    removeErrorMessage(decimalsErrorMessage);
```
daywiss, line 39: you could extract a lot of this logic into utility functions. not super necessary but then the logic could be tested independently

daywiss (approved):

hooks/helpers/useOnChange.ts
ryanwolhuter: agreed, will update

hooks/helpers/useOnChange.ts
ryanwolhuter: good idea, will do

PR #30, opened 2022-09-29
feat: show vote rolled

ryanwolhuter:
### Summary

We have a function that gets transaction hashes for votes from the chain. When there is no transaction hash available, we know that particular vote was rolled. Add an indicator in the UI that shows when a vote was rolled.

* Add an `isRolled` property to vote metadata
* Add a "rolled" indicator to the votes in the votes table


md0x (approved):
LGTM

PR #35, opened 2022-10-05
feat: create multiple error contexts to separate panel from other errors

daywiss:

This changes the error context to handle "typed" errors. These types are arbitrary and just specified with a string. Panels are updated to use their own types, to keep error messages separate between panels. This can easily be changed to share errors between panels if needed.


pages/api/augment-request.ts
```diff
@@ -69,7 +69,7 @@ async function getRequestTxFromL1RequestInformation(
   l1Requests.forEach((l1Request) => {
     let foundOoTx = { requestTransactionHash: "", chainId: 0, oracleType: "" };
     chainOoRequests.forEach((_, ooChainId) => {
-      chainOoRequests[ooChainId].forEach((ooRequest, index) => {
+      chainOoRequests[ooChainId].forEach((ooRequest) => {
```
daywiss, line 72: this is unrelated but i could not build, due to linting error. dont know how this passed ci in previous pr

pages/api/_common.ts
```diff
@@ -1,4 +1,4 @@
-import { ethers, Contract, BigNumber } from "ethers";
+import { ethers, Contract } from "ethers";
```
daywiss, line 1: this is unrelated but i could not build, due to linting error. dont know how this passed ci in previous pr

pages/_app.tsx
```diff
@@ -18,27 +19,29 @@ const queryClient = new QueryClient();
...
-                </BalancesProvider>
-              </ContractsProvider>
-              <ReactQueryDevtools />
-            </UserProvider>
-          </QueryClientProvider>
-        </WalletProvider>
-      </VoteTimingProvider>
-    </ErrorProvider>
+    <DefaultError.Provider>
+      <PanelError.Provider>
```
daywiss, line 23: 2 error contexts, one for panel, one for everything else

hooks/contexts/useErrorContext.ts
```diff
@@ -1,4 +1,13 @@
-import { ErrorContext } from "contexts";
+import { DefaultError, PanelError } from "contexts";
 import { useContext } from "react";
 
-export const useErrorContext = () => useContext(ErrorContext);
+export const useErrorContext = (type?:string) => {
```
daywiss, line 4: we can add arbitrary new error contexts here.  @ryanwolhuter im not sure if i should be dynamically calling useContext like this though, rather than having a static variable

components/PanelErrorBanner/PanelErrorBanner.tsx
```diff
@@ -3,7 +3,7 @@ import Warning from "public/assets/icons/warning.svg";
 import styled from "styled-components";
 
 export function PanelErrorBanner() {
-  const { errorMessages } = useErrorContext();
+  const { errorMessages } = useErrorContext("panel");
```
daywiss, line 6: this is how we access different contexts, if you dont provide a label, it will just use the default to not break compatibility with previous code

README.md
```diff
@@ -2,13 +2,23 @@ This is a [Next.js](https://nextjs.org/) project bootstrapped with [`create-next
 
 ## Getting Started
 
-First, run the development server:
+**Getting your environment variables**
+You can only do this if you have access to UMA's vercel projects.
+
+You need to install [vercel-cli](https://vercel.com/docs/cli).
+
+Then run `vercel link`, and follow the prompts to link the UMA->voter-dapp-v2 project.
```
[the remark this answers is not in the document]
ryanwolhuter: Ah thanks for this, I should have added it ages ago!

pages/api/augment-request.ts
ryanwolhuter: This now fixed 👍

pages/api/_common.ts
ryanwolhuter: This now fixed 👍

contexts/ErrorContext.tsx
```diff
@@ -14,33 +14,45 @@ export const defaultErrorContextState: ErrorContextState = {
...
-  }
-
-  function removeErrorMessage(message: ReactNode) {
-    setErrorMessages((prev) => prev.filter((prevMessage) => prevMessage !== message));
-  }
-
-  return (
-    <ErrorContext.Provider
+// This factory allows us to unstance multiple error contexts for different parts of the app.
+export function Factory(){
```
ryanwolhuter, line 18: I see a formatting problem here with `Factory(){` 

maybe run `yarn lint --fix`

hooks/contexts/useErrorContext.ts
ryanwolhuter: maybe it will be simpler to just have a `useErrorContext` and `usePanelErrorContext` to avoid this? I think this is cool but I don't know what the implications might be down the road.

daywiss: good idea

contexts/ErrorContext.tsx
```diff
@@ -14,33 +14,44 @@ export const defaultErrorContextState: ErrorContextState = {
...
-        errorMessages,
-        addErrorMessage,
-        removeErrorMessage,
-        clearErrorMessages,
-      }}
-    >
-      {children}
-    </ErrorContext.Provider>
-  );
+// This factory allows us to unstance multiple error contexts for different parts of the app.
```
ryanwolhuter, line 17: nit: spelling error "unstance" -> "instance"

package.json
```diff
@@ -7,6 +7,7 @@
     "build": "next build",
     "start": "next start",
     "lint": "next lint",
+    "lint-fix": "next lint --fix",
```
ryanwolhuter, line 10: +1

pages/_app.tsx
ryanwolhuter: nit: can we update this to use the same import pattern as the other contexts just for consistency?

ryanwolhuter (commented):
I like how this looks! One problem though — from the bug ticket (not sure if we all agree but lets just assume that we do) it says that they want to isolate errors to particular panels. So like for example now if an error happens while staking, and you close the stake panel and open the claim panel instead, you'll see the same error message there too. 

Can we add logic to filter the error messages so that we can filter by panel? Maybe change the error message from a string to an object like `{message: "Error!", origin: "stake-panel"}` or similar. What do you think?

daywiss:
> I like how this looks! One problem though — from the bug ticket (not sure if we all agree but lets just assume that we do) it says that they want to isolate errors to particular panels. So like for example now if an error happens while staking, and you close the stake panel and open the claim panel instead, you'll see the same error message there too.
> 
> Can we add logic to filter the error messages so that we can filter by panel? Maybe change the error message from a string to an object like {message: "Error!", origin: "stake-panel"} or similar. What do you think?

Ok good to know, i didnt realize that we needed to isolate errors between different panels. initially i started with what you are suggesting but found this method required less changes. ill see what i can do

components/Input/TextInput.tsx
```diff
@@ -24,7 +24,7 @@ export function TextInput({
   const inputMode = type === "text" ? "text" : type === "number" ? "decimal" : "email";
   // treat numbers as text inputs
   const _type = type === "number" ? "text" : type;
-  const onChange = useOnChange(onInput, maxDecimals, allowNegative);
+  const onChange = useOnChange(onInput, maxDecimals, allowNegative, _type);
```
daywiss, line 27: this was a bug i think, previously you could not edit the input box because the type would be defaulting to number

ryanwolhuter: awesome thanks for catching this

contexts/ErrorContext.tsx
```diff
@@ -1,34 +1,44 @@
 import { createContext, ReactNode, useState } from "react";
 
 export interface ErrorContextState {
-  errorMessages: ReactNode[];
-  addErrorMessage: (message: ReactNode) => void;
-  removeErrorMessage: (message: ReactNode) => void;
-  clearErrorMessages: () => void;
+  errorMessages: Record<string,ReactNode[]>;
```
ryanwolhuter, line 4: nit: formatting in `string,ReactNode` will make the build fail. I've fixed this issue with prettier not applying in a pending PR

contexts/ErrorContext.tsx
```diff
@@ -1,34 +1,44 @@
...
 export const ErrorContext = createContext<ErrorContextState>(defaultErrorContextState);
 
-export function ErrorProvider({ children }: { children: ReactNode }) {
-  const [errorMessages, setErrorMessages] = useState<ReactNode[]>([]);
+export function ErrorProvider ({ children }: { children: ReactNode }) {
+  const [errorMessages, setErrorMessages] = useState<Record<string,ReactNode[]>>({});
 
-  function addErrorMessage(message: ReactNode) {
-    setErrorMessages((prev) => [...new Set([...prev, message])]);
+  function addErrorMessage(type:string, message: ReactNode) {
```
ryanwolhuter, line 23: love it

ryanwolhuter (approved):
This is awesome, thank you so much.

The prettier issues are fixed in an open PR, but they're still coming through here. Maybe run a prettier -w manually just to sort them out.

PR #38, opened 2022-10-07
fix: reconfigure linting

ryanwolhuter:
### Summary

There were some problems with the eslint + prettier + typescript + nextjs buildchain. The main one being that nextjs does not lint all directories by default, only some default ones and then ones defined in `next.config.js`. While I was sorting this out I discovered a really cool extra feature of [typescript eslint](https://typescript-eslint.io/docs/linting/typed-linting) which adds a ton of extra useful typescript-specific linting rules. I added it here as well.

* Updated configs to lint all dirs
* Add and configure linting with type information
* Add workaround for svg imports being treated as type `any`


.eslintrc.json
```diff
@@ -1,33 +1,26 @@
...
-        "@typescript-eslint/no-unused-vars": [
-          "error",
-          { "ignoreRestSiblings": true }
-        ],
-        "@typescript-eslint/ban-ts-comment": 0
+  "extends": [
+    "next",
+    "plugin:@typescript-eslint/eslint-recommended",
+    "plugin:@typescript-eslint/recommended",
+    "plugin:@typescript-eslint/recommended-requiring-type-checking",
```
ryanwolhuter, line 11: While I was sorting out prettier I came across [this](https://typescript-eslint.io/docs/linting/typed-linting) 

It adds a ton of cool new rules that I wish I had before!

components/Panel/VotePanel/VotePanel.tsx
```diff
@@ -25,7 +25,7 @@ export function VotePanel({ content }: Props) {
     discordLink,
     participation,
     results,
-  } = content as VotePanelContentT;
+  } = content;
```
ryanwolhuter, line 28: removed redundant type cast

components/Toggle/Toggle.tsx
```diff
@@ -1,4 +1,4 @@
-import { blackOpacity25, green, red500 } from "constants/colors";
+import { blackOpacity25, red500 } from "constants/colors";
```
ryanwolhuter, line 1: removed unused var

components/Wallet/Wallet.tsx
```diff
@@ -29,10 +29,10 @@ export function Wallet() {
   useEffect(() => {
     if (!connect) return;
 
-    const previousConnectedWallets = JSON.parse(window.localStorage.getItem("connectedWallets") || "[]");
+    const previousConnectedWallets = JSON.parse(window.localStorage.getItem("connectedWallets") || "[]") as string[];
```
ryanwolhuter, line 32: added typecast

components/Wallet/Wallet.tsx
```diff
@@ -29,10 +29,10 @@ export function Wallet() {
...
 
     if (previousConnectedWallets?.length) {
-      (async () => {
+      void (async () => {
```
ryanwolhuter, line 35: New plugin suggests using `void` here to indicate that this function is not actually meant to return a promise even though it's marked as `async`. We get better typings like this.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void

helpers/decodeHexString.ts
```diff
@@ -10,6 +10,7 @@ export function decodeHexString(hexString: string) {
     if (e instanceof Error) {
       throw new Error(`Invalid hex string: ${e.message}`);
     } else {
```
ryanwolhuter, line 12: I don't see any other way than to ignore the rule here — we _want_ to attempt to stringify the error even if it's of type `unknown`.

helpers/formatVotes.ts
```diff
@@ -63,7 +63,7 @@ export async function formatVotesToCommit({
 
   return formattedVotes.filter((vote): vote is VoteFormattedToCommitT => Boolean(vote));
 }
-export async function formatVotesToReveal(decryptedVotesForUser: VoteT[]) {
+export function formatVotesToReveal(decryptedVotesForUser: VoteT[]) {
```
ryanwolhuter, line 66: removed redundant `async`

helpers/initOnboard.ts
```diff
@@ -8,8 +8,8 @@ const injected = injectedModule();
 const walletConnect = walletConnectModule();
 const gnosis = gnosisModule();
 
-const blocknativeDappId = process.env.NEXT_PUBLIC_BLOCKNATIVE_DAPP_ID;
-const infuraId = process.env.NEXT_PUBLIC_INFURA_ID;
+const blocknativeDappId = process.env.NEXT_PUBLIC_BLOCKNATIVE_DAPP_ID ?? "";
```
ryanwolhuter, line 11: add defaults

hooks/queries/votes/useDecryptedVotes.ts
```diff
@@ -33,7 +33,7 @@ async function decryptVotes(privateKey: string | undefined, encryptedVotes: Encr
 
     if (encryptedVote && privateKey) {
       const decryptedVoteString = await decryptMessage(privateKey, encryptedVote);
-      decryptedVote = JSON.parse(decryptedVoteString);
+      decryptedVote = JSON.parse(decryptedVoteString) as DecryptedVoteT;
```
ryanwolhuter, line 36: add typecast

next.config.js
```diff
@@ -2,9 +2,27 @@
 const nextConfig = {
   reactStrictMode: true,
   compiler: { styledComponents: true },
+  eslint: {
```
ryanwolhuter, line 5: This is the root of the problem with eslint / prettier not picking stuff up. Nextjs only lints certain dirs by default. Added other dirs here and now it catches the missing issues.

pages/api/_common.ts
```diff
@@ -1,4 +1,11 @@
-import { ethers, Contract } from "ethers";
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
```
ryanwolhuter, line 1: We can't avoid breaking these rules here because of the limitations of the types available to us.

pages/api/augment-request.ts
```diff
@@ -1,7 +1,14 @@
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
```
ryanwolhuter, line 1: We can't avoid breaking these rules here because of the limitations of the types available to us.

pages/api/decode-admin.ts
```diff
@@ -1,6 +1,13 @@
...
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+/* eslint-disable @typescript-eslint/no-unsafe-call */
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+/* eslint-disable @typescript-eslint/no-unsafe-return */
+/* eslint-disable @typescript-eslint/no-unsafe-argument */
+/* eslint-disable @typescript-eslint/no-explicit-any */
 import { NextApiRequest, NextApiResponse } from "next";
 
-// @ts-ignore
+// @ts-expect-error - no types for this module
```
ryanwolhuter, line 10: `@ts-ignore` is best avoided because it can mess up the compiler

types/images.d.ts
```diff
@@ -0,0 +1,46 @@
...
+  width: number;
+  placeholder?: string;
+};
+
+declare module "*.png" {
+  const content: StaticImageData;
+  export default content;
+}
+
+declare module "*.svg" {
```
ryanwolhuter, line 13: To make a very long story short, we need to override the definition for `.svg` files because nextjs treats them as `any` by default. This is wrong for us, because we use `svgr` to import and use svgs as react components.

web3/mutations/votes/revealVotes.ts
```diff
@@ -2,7 +2,7 @@ import { formatVotesToReveal } from "helpers";
 import { RevealVotes } from "types";
 
 export default async function revealVotes({ votesToReveal, voting }: RevealVotes) {
-  const formattedVotes = await formatVotesToReveal(votesToReveal);
+  const formattedVotes = formatVotesToReveal(votesToReveal);
```
ryanwolhuter, line 5: remove redundant await

components/Wallet/Wallet.tsx
daywiss: interesting, never seen this

contexts/ErrorContext.tsx
```diff
@@ -9,9 +9,9 @@ export interface ErrorContextState {
 
 export const defaultErrorContextState: ErrorContextState = {
   errorMessages: [],
-  addErrorMessage: () => {},
-  removeErrorMessage: () => {},
-  clearErrorMessages: () => {},
+  addErrorMessage: () => null,
```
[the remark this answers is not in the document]
daywiss: u might be also able to return void here

helpers/initOnboard.ts
```diff
@@ -8,8 +8,8 @@ const injected = injectedModule();
...
+const infuraId = process.env.NEXT_PUBLIC_INFURA_ID ?? "";
```
daywiss, line 12: if these are required for the app to run, id rather assert they exist and have the app fail with an error directing the dev to add the right envs. this makes it a lot easier to debug.  if these arent required then u can ignore this comment

daywiss (approved):
nice

PR #40, opened 2022-10-10
fix: dynamically fetch unstake time from contract, update copy

daywiss:

This fetches the cooldown time from the contract and updates relevant places in code to reflect time.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ yarn-error.log*
 
 # local env files
 .env*.local
+.env
 
 # vercel
 .vercel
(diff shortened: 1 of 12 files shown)


.gitignore
```diff
@@ -27,6 +27,7 @@ yarn-error.log*
 
 # local env files
 .env*.local
+.env
```
ryanwolhuter, line 30: What happened here? are we using another env file?

ryanwolhuter (approved):
awesome work and thank you for following the patterns perfectly!

PR #42, opened 2022-10-10
fix: use max approval when setting stake

daywiss:

There was an issue where staking multiple times wihtout wiating for previous transactions to go through would cause transactions to revert. this was because approvals were only set for the stake amount.  this sets unlimited approvals to prevent those reverts and prevent user from needing to re-approve. 
image


ryanwolhuter (approved):
lgtm thanks man!

PR #43, opened 2022-10-12
fix: claim button now visible

daywiss:

claim button was previously hidden when you had tokens to claim
image

this enables the button and lets user know why they cant unstake more if they have unclaimed tokens


ryanwolhuter (approved):
Awesome work as always, thanks David!

PR #44, opened 2022-10-14
feat: add delegation

ryanwolhuter:
### Summary

I am still busy polishing, styling, and testing this feature. However since it is so huge and complex, I figured I'd expedite getting the PR to ya'll so that we can all begin QA so long. 

The logic mostly is done (unknown bigs notwithstanding). I'll still need to add some polling logic to detect new requests as they come in.

To access this flow, connect your wallet and open the hamburger menu. Click the "wallet settings" link to go to the page where everything goes down.

@smb2796 I requested your review as well so you can test out the flow and start thinking about the copy. 

### Details

 I've set things up so that initiating a delegator-delegate relationship can only be done in one direction in the UI — one can send a request to another wallet to be their delegate, but not the other way around. This greatly reduces the number of states we need to account for. Also in my opinion this will be much more comprehensible to the user.
 
 The actions the user can do are:

1. send request be to be delegate — you are asking another wallet to be your delegate. This calls `setDelegate(otherWallet)`

2. cancel sent request to be delegate — undoes the above by calling `setDelegate(zeroAddress)`

accept received request to be delegate — completes the two-way relationship from the other wallet by calling `setDelegator(addressRequestCameFrom)`

ignore request to be delegate — if someone you don't want requests you to be their delegate, there's no way to stop them through the contracts. Instead we add the unwanted requester's address to a list of ignored addresses in local storage and don't show pending requests from there.

terminate relationship with delegate — breaks the two-way relationship from the delegator's side by `calling setDelegate(zeroAddress)`

terminate relationship with delegator — breaks the two-way relationship from the delegate's side by calling `setDelegator(zeroAddress)`


.storybook/.babelrc
```diff
@@ -1,4 +1,4 @@
 {
   "presets": ["@babel/preset-typescript", "@babel/preset-env", "@babel/preset-react"],
-  "plugins": ["react-require"]
+  "plugins": ["react-require", "babel-plugin-styled-components"]
```
ryanwolhuter, line 3: Wish I knew this sooner —  adding the babel plugin here lets us have the nice component names in the rendered classes in storybook too (previously this only worked on the dev server)

.storybook/preview.tsx
```diff
@@ -15,13 +16,30 @@ export const parameters = {
...
-    <QueryClientProvider client={new QueryClient()}>
-      <PanelProvider>
-        <GlobalStyle />
-        <Story />
-      </PanelProvider>
-    </QueryClientProvider>
-  </WalletProvider>
+  <ErrorProvider>
+      <VoteTimingProvider>
+        <WalletProvider>
```
ryanwolhuter, line 24: Updated the storybook preview with the new contexts that have been added recently. The preview should always mirror the `_app.tsx` file

components/Banner/Banner.tsx
```diff
@@ -1,12 +1,14 @@
+import { ReactNode } from "react";
 import styled from "styled-components";
 
-export function Banner() {
+interface Props {
```
ryanwolhuter, line 4: This component gets used in other places now so I made it generic

components/Banner/Banner.tsx
```diff
@@ -18,7 +20,7 @@ const OuterWrapper = styled.div`
 
 const InnerWrapper = styled.div`
   max-width: var(--desktop-max-width);
-  height: 125px;
+  height: var(--banner-height);
```
ryanwolhuter, line 23: Use a global var so that this value can be used in calculations elsewhere (we'll also need this for the mobile views soon)

components/GlobalStyle/GlobalStyle.ts
```diff
@@ -76,6 +76,10 @@ html {
   }
 }
 
+html, body, #__next {
```
ryanwolhuter, line 79: I thought I had included this already but it seems not. This lets us do `height:100%` to take up the full height of the viewport.

components/Header/Header.tsx
```diff
@@ -42,7 +42,7 @@ const OuterWrapper = styled.header``;
 
 const InnerWrapper = styled.div`
   max-width: var(--desktop-max-width);
-  min-height: 80px;
+  min-height: var(--header-height);
```
ryanwolhuter, line 45: Again switched to var to use in calculations

components/Layout/Layout.tsx
```diff
@@ -1,16 +1,20 @@
...
       <Header />
-      <main>{children}</main>
-      <Panel />
-    </>
+      {children}
+    </Main>
   );
 }
+
+const Main = styled.main`
```
ryanwolhuter, line 18: This is also required for doing `height:100%` to take up the full height of the view.

components/Nav/Nav.tsx
```diff
@@ -11,10 +12,20 @@ interface Props {
...
 }
 export function Nav({ links }: Props) {
-  const { pathname } = useRouter();
-  const isActive = (href: string) => pathname === href;
+  const { closePanel } = usePanelContext();
+  const router = useRouter();
+  const isActive = (href: string) => router.pathname === href;
   const isExternalLink = (href: string) => !href.startsWith("/");
 
+  useEffect(() => {
```
ryanwolhuter, line 20: Since we have nav links in the panel, navigation looks a lot cleaner when we close the panel on navigation.

components/Panel/ClaimPanel.tsx
```diff
@@ -10,7 +10,7 @@ export function ClaimPanel() {
   const { voting } = useContractsContext();
   const { withdrawRewardsMutation, isWithdrawingRewards } = useWithdrawRewards("claim");
   const { withdrawAndRestakeMutation, isWithdrawingAndRestaking } = useWithdrawAndRestake("claim");
-  const { outstandingRewards, getBalancesFetching } = useBalancesContext();
+  const { outstandingRewards, getStakingDataFetching } = useStakingContext();
```
ryanwolhuter, line 13: Just some renaming here, logic is the same

components/Panel/DelegationPanel.tsx
```diff
@@ -0,0 +1,236 @@
...
+import Three from "public/assets/icons/three.svg";
+import Time from "public/assets/icons/time-with-inner-circle.svg";
+import Two from "public/assets/icons/two.svg";
+import { useEffect, useState } from "react";
+import styled, { css } from "styled-components";
+import { PanelFooter } from "./PanelFooter";
+import { PanelTitle } from "./PanelTitle";
+import { PanelSectionText, PanelSectionTitle, PanelWrapper } from "./styles";
+
+export function DelegationPanel() {
```
ryanwolhuter, line 15: This panel shows when the user wants to add a delegate wallet.

components/Panel/DelegationPanel.tsx
```diff
@@ -0,0 +1,231 @@
...
+    sendRequestToBeDelegate,
+    getPendingSentRequestsToBeDelegate,
+    getDelegationDataFetching,
+  } = useDelegationContext();
+
+  const delegationStatus = getDelegationStatus();
+
+  useEffect(() => {
+    // only show this panel when the user has not yet entered a delegation relationship, or the user has requested another wallet to be their delegate wallet
+    if (!(delegationStatus === "no-delegation" || delegationStatus === "delegator-pending")) {
```
ryanwolhuter, line 31: The only time the user can show this panel is when clicking "add wallet" on the wallet settings page. This button is only shown when the user has not requested another wallet to be their delegate. Hence we close the panel in other cases. The exception is that when user completes this add wallet flow and moves into the `delegator-pending` status, we don't want to close the panel and instead show the user information about the request they just made.

components/Panel/DelegationPanel.tsx
```diff
@@ -0,0 +1,231 @@
...
+    setDelegateAddressToAdd(inputAddress);
+
+    if (shouldRunValidation(inputAddress)) {
+      if (!validateInputAddress(inputAddress)) return;
+    } else {
+      clearErrorMessages();
+    }
+  }
+
+  function validateInputAddress(inputAddress: string) {
```
ryanwolhuter, line 61: This validation logic is simple but I think it is sufficient. We check that the address is a legit address, and also that the user is not trying to delegate to themself.

components/Panel/DelegationPanel.tsx
```diff
@@ -0,0 +1,231 @@
...
+      return false;
+    }
+    if (getAddress(inputAddress) === getAddress(address)) {
+      addErrorMessage("You cannot delegate to yourself.");
+      return false;
+    }
+    return true;
+  }
+
+  function shouldRunValidation(address: string) {
```
ryanwolhuter, line 73: Since ethers supports addresses without the "0x" prefix, we decide at what character count to begin validation based on the presence of "0x"

components/Panel/MenuPanel/MenuPanel.tsx
```diff
@@ -1,62 +1,66 @@
...
+      title: "Optimistic Oracle",
+      href: "https://oracle.umaproject.org",
+    },
+    {
+      title: "Docs",
+      href: "https://docs.umaproject.org",
+    },
+  ];
+
+  const _links = connectedWallet ? links : links.filter((link) => link.href !== "/wallet-settings");
```
ryanwolhuter, line 37: We don't show the "wallet settings" link when there is no wallet connected to avoid confusion.

components/Panel/MenuPanel/MenuPanel.tsx
```diff
@@ -1,62 +1,66 @@
...
 
   return (
     <PanelWrapper>
       <AccountWrapper>
         <Title>Account</Title>
-        {connectedWallets.length ? (
+        {connectedWallet ? (
```
ryanwolhuter, line 43: Logic here is the same, just getting the data from the context now as is the pattern elsewhere

components/Votes/Votes.tsx
```diff
@@ -111,60 +105,40 @@ export function Votes() {
...
-                  phase={phase}
-                  selectedVote={selectedVotes[vote.uniqueKey]}
-                  selectVote={(value) => selectVote(value, vote)}
-                  activityStatus={getActivityStatus()}
-                  moreDetailsAction={() => openVotePanel(vote)}
-                  key={vote.uniqueKey}
-                  isFetching={getUserDependentIsFetching() || isCommittingVotes || isRevealingVotes}
-                />
-              ))}
+    <>
```
ryanwolhuter, line 108: Logic is the same here too. I just refactored this so that the voting and wallet settings pages have the same pattern for layout markup.

components/Wallet/Wallet.tsx
```diff
@@ -27,8 +26,7 @@ export function Wallet() {
   }, [connectedWallets, wallet]);
 
   useEffect(() => {
-    if (!connect) return;
-
+    if (!connect || connectedWallets?.length > 0) return;
```
ryanwolhuter, line 29: This was a bug. It would run the automatic connection logic even when a connected wallet existed, which would show an error

components/pages/Vote.tsx
```diff
@@ -1,11 +1,37 @@
...
+import { Banner, HowItWorks, Layout, LoadingSpinner, PageInnerWrapper, PageOuterWrapper, Votes } from "components";
+import { useVotesContext } from "hooks";
+import styled from "styled-components";
 
 export function Vote() {
+  const { getUserIndependentIsLoading } = useVotesContext();
   return (
     <Layout>
-      <Banner />
+      <Banner>
```
ryanwolhuter, line 9: Moved the markup for the layout here from the `Votes` component so that this pattern matches the one in `WalletSettings`

contexts/DelegationContext.tsx
```diff
@@ -0,0 +1,292 @@
...
+  useTerminateRelationshipWithDelegate,
+  useTerminateRelationshipWithDelegator,
+  useUserContext,
+  useVoterFromDelegate,
+} from "hooks";
+import { createContext, ReactNode } from "react";
+import { DelegationEventT, DelegationStatusT } from "types";
+export interface DelegationContextState {
+  getDelegationStatus: () => DelegationStatusT;
+  getPendingReceivedRequestsToBeDelegate: () => DelegationEventT[];
```
ryanwolhuter, line 25: These methods define the actions that the user can take from the frontend. I have deliberately limited these actions so that initiating a delegator-delegate relationship can only be done in one direction — one can send a request to another wallet to be their delegate, but not the other way around. This greatly reduces the number of states we need to account for. Also in my opinion this will be much more comprehensible to the user.

contexts/DelegationContext.tsx
```diff
@@ -0,0 +1,292 @@
...
+  getHasPendingReceivedRequestsToBeDelegate: () => boolean;
+  getPendingSentRequestsToBeDelegate: () => DelegationEventT[];
+  getHasPendingSentRequestsToBeDelegate: () => boolean;
+  getDelegateAddress: () => string | undefined;
+  getDelegatorAddress: () => string | undefined;
+  sendRequestToBeDelegate: (delegateAddress: string) => void;
```
ryanwolhuter, line 31: The actions the user can do are:

* send be to be delegate — you are asking another wallet to be your delegate. This calls `setDelegate(otherWallet)`
* cancel sent request to be delegate — undoes the above by calling `setDelegate(zeroAddress)
* accept received request to be delegate — completes the two-way relationship from the other wallet by calling `setDelegator(walletRequestCameFrom)`
* ignore request to be delegate — if someone you don't want requests you to be their delegate, there's no way to stop them through the contracts. Instead we add the unwanted requester's address to a list of ignored addresses in local storage and don't show pending requests from there.

contexts/StakingContext.tsx
```diff
@@ -9,7 +9,7 @@ import {
 } from "hooks";
 import { createContext, ReactNode } from "react";
 
-export interface BalancesContextState {
+export interface StakingContextState {
```
ryanwolhuter, line 12: I renamed this context because it also contains information about delegation, so I felt this name is more appropriate

hooks/mutations/delegation/useIgnoreReceivedRequestToBeDelegate.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+function ignoreReceivedRequestToBeDelegate({
+  userAddress,
+  delegatorAddress,
+}: {
+  userAddress: string;
+  delegatorAddress: string;
+}) {
+  return new Promise(() => {
```
ryanwolhuter, line 40: react query only works with functions that return promises, so this is wrapped in a promise even though it is not an async operation

daywiss: why not make the function `ignoreReceivedRequestToBeDelegate` async. also i dont think this promise works right because its never resolved or rejected

ryanwolhuter: I actually did it like that at first but then the typescript eslint complained. But yeah thinking of it again I prefer just making it async and I'll add an ignore there.

components/Panel/DelegationPanel.tsx
```diff
@@ -0,0 +1,231 @@
...
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [delegationStatus]);
+
+  const pendingRequests = getPendingSentRequestsToBeDelegate();
+
+  // only allow user to add a delegate wallet if they are neither a delegator nor a delegate
+  const showAddDelegateInput = delegationStatus === "no-delegation";
+
+  const showPendingRequests = delegationStatus === "delegator-pending" && pendingRequests.length > 0;
```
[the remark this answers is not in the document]
ryanwolhuter: yeah you're right will add

components/pages/WalletSettings/PendingRequests.tsx
```diff
@@ -0,0 +1,100 @@
+import NextLink from "next/link";
+import Time from "public/assets/icons/time-with-inner-circle.svg";
+import styled, { CSSProperties } from "styled-components";
+import { DelegationEventT } from "types";
+import { AddressWrapper, BarButtonPrimary, BarButtonSecondary, BarWrapper, Header, Text } from "./styles";
+
+export function PendingRequests({
```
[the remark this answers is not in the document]
ryanwolhuter: I'm hoping to change all of these names to things that make much more sense once we have feedback on the terminology / copy

contexts/DelegationContext.tsx
```diff
@@ -0,0 +1,292 @@
...
+    // if there is a delegator set for your address, you are a delegate
+    if (voterFromDelegate.toLowerCase() !== address.toLowerCase()) return "delegate";
+    // if the `delegateToStaker` mapping for the `delegate` defined in your `voterStakes`, then you are a delegator
+    if (address.toLowerCase() === delegateToStaker.toLowerCase()) return "delegator";
+    // if the user has received a request to be another wallet's delegate but they have not accepted any, then they are a pending delegate
+    if (getHasPendingReceivedRequestsToBeDelegate()) return "delegate-pending";
+    // if the user has sent a request to be another wallet's delegate but the other wallet has not yet accepted, then they are a pending delegator
+    if (getHasPendingSentRequestsToBeDelegate()) return "delegator-pending";
+    // if none are true we assume the user has no delegation
+    return "no-delegation";
```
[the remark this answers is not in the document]
ryanwolhuter: Yeah this is the intended behavior. It's not possible (from the UI) to send a request to be a _delegator_. You can only accept an _existing_ request to be a delegate. So the only states are either you have _sent_ or _received_ a request to be a delegate.

components/pages/WalletSettings/OtherWallet.tsx
```diff
@@ -0,0 +1,61 @@
...
+  if (!address) return null;
+
+  const text =
+    status === "delegate"
+      ? "Explanation of secondary wallet when secondary is delegate."
+      : "Explanation of secondary wallet when secondary is delegator.";
+
+  return (
+    <>
+      <Header>Other wallet ({status})</Header>
```
md0x, line 24: nit: this shows the status in lower case while `Connected Wallet (Delegate/Delegator)` has the first letter in upper case

contexts/DelegationContext.tsx
```diff
@@ -0,0 +1,292 @@
...
+              (delegatorSet) =>
+                delegatorSet.delegate == delegateSet.delegate && delegatorSet.delegator == delegatorSet.delegator
+            )
+        )
+        ?.filter(({ delegator }) => !ignoredRequestToBeDelegateAddresses?.includes(delegator)) ?? []
+    );
+  }
+
+  function getPendingSentRequestsToBeDelegate() {
+    return (
```
md0x, line 217: After doing a `setDelegate(0x0)` I'm still having "Pending delegator requests" 

If the "delegator" has set the delegate to `0x0`, ignore the pending delegator requests:

`if(delegate=0x0) return [];`

hooks/mutations/delegation/useCancelSentRequestToBeDelegate.ts
```diff
@@ -0,0 +1,25 @@
...
+import { useHandleError, useUserContext } from "hooks";
+import { DelegationEventT } from "types";
+import { removeDelegator } from "web3";
+
+export function useCancelSentRequestToBeDelegate() {
+  const { address } = useUserContext();
+  const onError = useHandleError();
+  const queryClient = useQueryClient();
+
+  const { mutate, isLoading } = useMutation(removeDelegator, {
```
md0x, line 12:
```suggestion
  const { mutate, isLoading } = useMutation(removeDelegate, {
```

hooks/mutations/delegation/useSendRequestToBeDelegate.ts
```diff
@@ -0,0 +1,16 @@
+import { useMutation } from "@tanstack/react-query";
+import { useHandleError } from "hooks";
+import { setDelegate } from "web3";
+
+export function useSendRequestToBeDelegate(errorType?: string) {
```
md0x, line 5: Here the delegator is doing a setDelegate to eventually become a delegator right?
I think we should rename it:
```suggestion
export function useSendRequestToBeDelegator(errorType?: string) {
```

hooks/mutations/delegation/useCancelSentRequestToBeDelegate.ts
```diff
@@ -0,0 +1,25 @@
+import { useMutation, useQueryClient } from "@tanstack/react-query";
+import { sentRequestsToBeDelegateKey } from "constants/queryKeys";
+import { useHandleError, useUserContext } from "hooks";
+import { DelegationEventT } from "types";
+import { removeDelegator } from "web3";
+
+export function useCancelSentRequestToBeDelegate() {
```
md0x, line 7: This is used by a `pending-delegator` to cancel a pending setDelegate right?
I think it should renamed:
```suggestion
export function useCancelSentRequestToBeDelegator() {
```

hooks/mutations/delegation/useCancelSentRequestToBeDelegate.ts
```diff
@@ -0,0 +1,25 @@
+import { useMutation, useQueryClient } from "@tanstack/react-query";
+import { sentRequestsToBeDelegateKey } from "constants/queryKeys";
+import { useHandleError, useUserContext } from "hooks";
+import { DelegationEventT } from "types";
+import { removeDelegator } from "web3";
```
md0x, line 5:
```suggestion
import { removeDelegate } from "web3";
```

contexts/DelegationContext.tsx
```diff
@@ -0,0 +1,292 @@
...
+      sentRequestsToBeDelegate?.filter(
+        (delegateSet) =>
```
md0x, line 219:
```suggestion
        (delegateSet) =>
        delegateSet.delegate !== ADDRESS_ZERO && // filter out the cancel delegations events
```

hooks/mutations/delegation/useSendRequestToBeDelegate.ts
ryanwolhuter: What's happening here is that the user is asking someone else to be their delegate. So we could phrase it either as "asking someone to be your delegate" OR "asking someone to be their delegator" and either would be correct. I've chosen the former way here, but I'm reaaaally hoping that with some input from @smb2796 et al we'll change the terms here and we can do something a bit less confusing.

components/Panel/DelegationPanel.tsx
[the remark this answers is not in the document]
daywiss: i think the only real benefit of having them defined in a constant is having a place to see all the options and document them, but yea i typically try to avoid defining them if i can. it just seems like this is used in a lot of places and theres like 5 different statuses to know. not super opinionated but i think it would help when someone else needs to debug or add feature

md0x:
NIT:
I think we could only have the spinner in the right panel and maybe have a placeholder in the "other wallet" component
image

components/Panel/DelegationPanel.tsx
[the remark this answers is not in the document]
md0x: I agree, it could be nice to define all the wallet status in an `enum WalletDelegationTypes{...}`

Some pros:
- readability: helps understand for example what a `"no-delegation"` is: it's a wallet delegation type or status
- maintainability: helps a new dev to understand the different types of delegation, modify them, fix them, etc... command clicking on the type and seeing the rest of types is helpful I think
- repetition: less repetition

Generally, I think that grouping elements/values of the same type of thing is useful. What do you think?

hooks/mutations/delegation/useSendRequestToBeDelegate.ts
smb2796: So on the naming convention - agreed that using delegator and delegate is pretty confusing (especially from code perspective). I looked at Aave, ENS, Uniswap etc, and they all use delegate / delegator, but do not have this explicit acceptance from delegate side, so it does not confuse UX as much. A few options:
- Delegator & Representative
- Delegator and Voter
- Owner and delegate
- Tokenholder and delegate

My preference is probably the first because I think it works for all case (setTokenholder doesn't really make too much sense), but perhaps it is too long. Thoughts?

components/Panel/DelegationPanel.tsx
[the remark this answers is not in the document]
ryanwolhuter: The delegation status is defined by the type:

`
export type DelegationStatusT =
  | "no-wallet-connected"
  | "no-delegation"
  | "delegate"
  | "delegator"
  | "delegate-pending"
  | "delegator-pending";
  `
  
  The delegation status can only be one of these strings.
  
  To me this is:
  
  * more readable — `status === "no-wallet-connected"` vs `status === DelegationStatuses.NoWalletConnected`
  * more maintainable — no need to `import DelegationStatuses` and also editor autocomplete fills in the strings / takes you to the type definition
  * less repetition — with an enum you have to write `DelegationStatuses.` every time.
  
  Just my thoughts

hooks/mutations/delegation/useSendRequestToBeDelegate.ts
ryanwolhuter: Interesting that the others use delegator / delegate. Now that I've gotten more used to the terms, they've sort of grown on me. Maybe they aren't that confusing after all.

I really am not sure what I think there. 

@chrismaree @md0x @mrice32 do you guys maybe wanna weigh in here?

daywiss:
https://linear.app/uma/issue/UMA-633/tx-link-generation-is-wrong

daywiss:
https://linear.app/uma/issue/UMA-635/removing-delegate-behaves-badly

hooks/mutations/delegation/useSendRequestToBeDelegate.ts
smb2796: As long as it's not too confusing from code perspective, I don't think it is for UX

chrismaree (approved):

PR #45, opened 2022-10-21
fix: format ethers tx errors for user

daywiss:
Transaction errors now get formatted to only include the message part. previously it would include call data.
image

This may not cover all transactions, so please comment if there needs to be more covered.

### DIFF
--- a/components/PanelErrorBanner/PanelErrorBanner.tsx
+++ b/components/PanelErrorBanner/PanelErrorBanner.tsx
@@ -42,4 +42,7 @@ const ErrorMessage = styled.p`
   }
 `;
 
-const IconWrapper = styled.div``;
+const IconWrapper = styled.div`
+  width: 26px;
+  height: 26px;
+`;
--- a/helpers/ethers.ts
+++ b/helpers/ethers.ts
@@ -1,4 +1,5 @@
 import { ethers } from "ethers";
+import { capitalizeFirstLetter } from "./misc";
 
 export const formatEther = ethers.utils.formatEther;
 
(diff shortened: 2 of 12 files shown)


components/PanelErrorBanner/PanelErrorBanner.tsx
```diff
@@ -37,9 +37,13 @@ const InnerWrapper = styled.div``;
 
 const ErrorMessage = styled.p`
   overflow-wrap: anywhere;
+  text-transform: capitalize;
```
ryanwolhuter, line 40: why are we capitalizing this? its just a regular sentence, not a title

helpers/ethers.ts
```diff
@@ -13,3 +13,11 @@ export const toUtf8String = ethers.utils.toUtf8String;
 export const formatBytes32String = ethers.utils.formatBytes32String;
 
 export const commify = ethers.utils.commify;
+
+export const formatTransactionError = (error: unknown): string | unknown => {
```
ryanwolhuter, line 17: 🔥

helpers/index.ts
```diff
@@ -11,15 +11,7 @@ export {
...
-export {
-  commify,
-  formatBytes32String,
-  formatEther,
-  parseEther,
-  randomBytes,
-  solidityKeccak256,
-  toUtf8String,
-} from "./ethers";
+export * from "./ethers";
```
ryanwolhuter, line 14: should have done it like this from the start

ryanwolhuter (approved):
great work

components/PanelErrorBanner/PanelErrorBanner.tsx
daywiss: i actually just want to capiatlize the first letter of the first word because it makes the default ethers errors look better. i can remove this though and do it in the string formatting

helpers/index.ts
daywiss: actaully im going to revert this because it makes it harder to discover the functions in each file when goign to the index page

PR #46, opened 2022-10-24
fix: catch ether parsing errors from user input

daywiss:

This adds a second ether parsing function which catches an error and returns 0. A better but more work solution would be to validate user input before passing it to be parsed, but I don't think the extra work is justified, as this should work fine.


ryanwolhuter (approved):
awesome

PR #47, opened 2022-10-24
fix: have button links detect when to open new tab

daywiss:

Takes external link detection, puts it as a utility and uses it in a link button. This should fix the issue, but I wasnt able to identify the particular local in the bug report.


ryanwolhuter (approved):
lgtm

PR #48, opened 2022-10-24
fix: include pending unstake in total balance

daywiss:

We should be including any pending unstake amount in the total balance


ryanwolhuter (approved):
👍

PR #50, opened 2022-10-24
fix: dont let staking balance go below 0

daywiss:

Looks like there is a bit of a race condition when doing this subtraction. When unstaking we somtimes see an negative value, this is due to the state of staked balance changing before we do the subtraction.  we just make sure here to only show 0 if the number goes negative.


ryanwolhuter (approved):
👍

PR #53, opened 2022-10-28
feat: add tooltips and notifications

ryanwolhuter:
### Summary

Adds two similar components — one for showing tooltips, and another for showing notifications. Also adds a context for controlling notifications, and implements the notification-showing logic in the contract interaction mutations.

See comments in the PR for implementation details.


components/Notifications/Notification.tsx
```diff
@@ -0,0 +1,113 @@
+import { LoadingSpinner } from "components/LoadingSpinner";
+import NextLink from "next/link";
+import Check from "public/assets/icons/check.svg";
+import Close from "public/assets/icons/close.svg";
+import Warning from "public/assets/icons/warning.svg";
+import styled from "styled-components";
+import { NotificationT } from "types";
+
+export function Notification({
```
ryanwolhuter, line 9: This is a presentational component that shows a contract interaction notification. It handles three types: "success" | "error" | "pending"

components/Notifications/Notifications.tsx
```diff
@@ -0,0 +1,39 @@
+import { Portal } from "@reach/portal";
```
ryanwolhuter, line 1: a `Portal` is a component that is appended to the end of the document body. This lets us make sure that it sits above everything else without needing to engage in z index wars. The `Panel` works in the same way.

components/Notifications/Notifications.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { animated, useTransition } from "@react-spring/web";
+import { useNotificationsContext } from "hooks";
+import styled from "styled-components";
+import { Notification } from "./Notification";
+
+export function Notifications() {
+  const { notifications, removeNotification } = useNotificationsContext();
+
+  const transitions = useTransition(notifications, {
```
ryanwolhuter, line 10: I've used the same super basic enter/leave animation for the notifications and tooltips for now. its definitely good enough but it can be improved later.

components/Notifications/Notifications.tsx
```diff
@@ -0,0 +1,39 @@
...
+        )}
+      </Wrapper>
+    </Portal>
+  );
+}
+
+const AnimatedNotification = animated(Notification);
+
+const Wrapper = styled.div`
+  z-index: 1;
```
ryanwolhuter, line 32: We still need a z index because this must be placed above the `Panel` (also a portal) but you'll see that we only need it to be 1.

components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,40 @@
...
+import "@reach/tooltip/styles.css";
+import { animated, useTransition } from "@react-spring/web";
+import { cloneElement } from "react";
+import styled from "styled-components";
+
+interface Props {
+  "aria-label"?: string;
+  children: React.ReactElement;
+  label: string;
+}
```
ryanwolhuter, line 11: A simple tooltip component. The child element passed to it will trigger it to show when hovered.

components/Panel/ClaimPanel.tsx
```diff
@@ -18,14 +19,43 @@ export function ClaimPanel() {
...
   const { withdrawAndRestakeMutation, isWithdrawingAndRestaking } = useWithdrawAndRestake("claim");
   const { outstandingRewards, getStakingDataFetching } = useStakingContext();
+  const notificationHandler = useContractInteractionNotifications();
   const isDelegate = getDelegationStatus() === "delegate";
 
   function withdrawRewards() {
-    withdrawRewardsMutation({ voting });
+    if (!outstandingRewards) return;
+
+    withdrawRewardsMutation(
```
ryanwolhuter, line 28: this is the pattern used for most of the mutations. you'll notice that he use `notificationHandler` in two different situations: for showing pending transactions (uses a `ContractTransaction`) and for showing settled (succeeded or failed) which uses a `ContractReceipt`.

We must pass the notification handler to the mutation's getter function because otherwise we would have to wait for the transaction to resolve before we would have access to the transaction hash (see the implementation in the getter functions for more details.)

components/VotesTable/VotesTableRow.tsx
```diff
@@ -148,15 +148,17 @@ export function VotesTableRow({
...
-                <RolledWrapper>
-                  <RolledIconWrapper>
-                    <RolledIcon />
-                  </RolledIconWrapper>
-                  {/* todo: add link to explanation of rolled votes in the docs once its written */}
-                  <Link href="https://docs.umaproject.org" passHref>
-                    <RolledLink target="_blank">Rolled</RolledLink>
-                  </Link>
-                </RolledWrapper>
+                <Tooltip label="This vote was included in the previous voting cycle, but did not get enough votes to resolve.">
```
ryanwolhuter, line 151: Added an example usage of the tooltip here. We will add more soon.

contexts/DelegationContext.tsx
```diff
@@ -234,40 +236,108 @@ export function DelegationProvider({ children }: { children: ReactNode }) {
       {
         voting,
         delegateAddress,
+        notificationMessage: `Requesting ${truncateEthAddress(delegateAddress)} to be your delegate...`,
```
ryanwolhuter, line 239: The logic for the delegation mutations is slightly more complicated because we need more information about what's happening to create the message. The same mutations are used in different contexts, so we must specify the context when calling the function.

contexts/NotificationsContext.tsx
```diff
@@ -0,0 +1,88 @@
+import { createContext, ReactNode, useState } from "react";
+import { AddNotificationFn, NotificationT, RemoveNotificationFn } from "types";
+
+export interface NotificationsContextState {
```
ryanwolhuter, line 4: This context contains the notifications. It is similar to the error context.

contexts/NotificationsContext.tsx
```diff
@@ -0,0 +1,88 @@
...
+  function addSuccessNotification(description: ReactNode, transactionHash: string) {
+    addNotification({
+      description,
+      transactionHash,
+      type: "success",
+    });
+
+    setTimeout(() => {
+      removeNotification(transactionHash);
+    }, 5000);
```
ryanwolhuter, line 50: Autohide settled notifications. We don't need to auto-hide pending ones because they are removed once the mutation is settled.

helpers/initOnboard.ts
```diff
@@ -36,19 +36,7 @@ export const initOnboard = init({
...
-        return {
-          // autoDismiss set to zero will persist the notification until the user excuses it
-          autoDismiss: 0,
-          // message: `Your transaction is pending, click <a href="https://etherscan.io/tx/${transaction.hash}" rel="noopener noreferrer" target="_blank">here</a> for more info.`,
-          // or you could use onClick for when someone clicks on the notification itself
-          onClick: () => window.open(`https://etherscan.io/tx/${transaction.hash}`),
-        };
-      }
-    },
+    enabled: false,
```
ryanwolhuter, line 39: We are not using notify anymore.

hooks/helpers/useContractInteractionNotifications.ts
```diff
@@ -0,0 +1,42 @@
+import { useNotificationsContext } from "hooks";
+import { ContractInteractionNotificationT } from "types";
+
+export function useContractInteractionNotifications() {
```
ryanwolhuter, line 4: This hook handles notifications for contract transactions and receipts. transactions are by definition pending, and receipts are settled.

components/Panel/ClaimPanel.tsx
daywiss: i understand why you did it this way, but its really annoying to have to pass this around everywhere you need it. since you are building up an api to call contracts outside of react, why not add a notification "dispatcher" outside of react as well and import it where you need it?  You could do this with a few lines by instantiating an event emitter singleton which could be pulled into any file.  In your notification context subscribe to events, and add notifications from it. In any other place in the app, import the event emitter and emit notification events when you need it.  Then you can effectively eliminate passing around the notification handler.  [Here is an example](https://github.com/UMAprotocol/optimistic-oracle-dapp/blob/master/src/context/OracleClientContext.tsx#L36)

daywiss (commented):
some potential improvements, let me know what u think

components/Notifications/Notifications.tsx
ryanwolhuter: It's the same library that's used to animate the `Panel` component. It let's us be aware of when a component is shown or hidden, and add enter/leave transitions to that. This particular library [`react-spring`](https://react-spring.dev/) is one of the most popular animation libraries going around. It's advantage is that it uses spring-based physics to produce more natural looking animations than what can be achieved with css transitions alone.

contexts/NotificationsContext.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  function addNotification(notification: NotificationT) {
+    setNotifications((prev) => [...prev, notification]);
+  }
+
+  function clearNotifications() {
+    setNotifications([]);
+  }
+
+  function removeNotification(transactionHash: string) {
```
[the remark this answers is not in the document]
ryanwolhuter: I had originally intended for this to be used exclusively for on chain interaction notifications, since we've pushed the other notification features to v2.1. I could make this more generic, but that would involve more work up front. I figured we might as well go with the minimal solution that works for now, and we can refactor to be more general when the time comes.

components/Panel/ClaimPanel.tsx
ryanwolhuter: this is an awesome solution and I'm excited to implement it. I'll re-request your review once its done.

daywiss (approved):
looks good! im stil a bit unsure of why you still need tx specific logic in the notifier, but its not a blocker.

types/global.ts
```diff
@@ -240,6 +242,23 @@ export type PageStateT = {
 
 export type PageStatesT = Record<PaginateForT, PageStateT>;
 
+export type UuidT = ReturnType<typeof uuid>;
+
+export type NotificationT = {
+  message: ReactNode;
+  id: UuidT;
+  transactionHash?: string;
```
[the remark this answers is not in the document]
ryanwolhuter: the transaction hash is optional — the message is a react node in case you need to include other markup in there like <strong> etc. The transaction hash is used to generate another piece of markup with an etherscan link, if it exists.

we are doing different templates based on the type — success is a check mark, error is a warning icon, and pending shows a loading spinner.

types/global.ts
```diff
@@ -1,5 +1,7 @@
 import { VotingV2Ethers } from "@uma/contracts-frontend";
 import { BigNumber } from "ethers";
+import { ReactNode } from "react";
+import uuid from "react-uuid";
```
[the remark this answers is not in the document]
ryanwolhuter: ohhh this is cool, yeah I just reached for the first thing I could find but this sounds better.

PR #56, opened 2022-11-02
feat: add transaction links to commit reveal txs

daywiss:

This adds transaction hash to the reveal/commit data when we get that data for the user. this is used to create a clickable button to go to etherscan when those exist. 

this is now a link:
image

### DIFF
--- a/hooks/mutations/votes/useCommitVotes.ts
+++ b/hooks/mutations/votes/useCommitVotes.ts
@@ -12,13 +12,16 @@ export function useCommitVotes() {
 
   const { mutate, isLoading } = useMutation(commitVotes, {
     onSuccess: (_data, { formattedVotes }) => {
+      console.log({ _data });
       queryClient.setQueryData<VoteExistsByKeyT>(
         [committedVotesKey, address, roundId],
         (oldCommittedVotes) => {
           const newCommittedVotes = { ...oldCommittedVotes };
 
           for (const { uniqueKey } of formattedVotes) {
-            newCommittedVotes[uniqueKey] = true;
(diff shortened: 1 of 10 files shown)


hooks/mutations/votes/useCommitVotes.ts
```diff
@@ -12,13 +12,16 @@ export function useCommitVotes() {
 
   const { mutate, isLoading } = useMutation(commitVotes, {
     onSuccess: (_data, { formattedVotes }) => {
+      console.log({ _data });
```
ryanwolhuter, line 15: stray log

hooks/mutations/votes/useCommitVotes.ts
```diff
@@ -12,13 +12,16 @@ export function useCommitVotes() {
...
       queryClient.setQueryData<VoteExistsByKeyT>(
         [committedVotesKey, address, roundId],
         (oldCommittedVotes) => {
           const newCommittedVotes = { ...oldCommittedVotes };
 
           for (const { uniqueKey } of formattedVotes) {
-            newCommittedVotes[uniqueKey] = true;
+            if (_data?.transactionHash) {
```
ryanwolhuter, line 22: the _ with data here makes it ignored (let's you define the param without using it) if you do want to use it, rather remove the _.

ryanwolhuter (approved):
great work, just some tiny nits

hooks/mutations/votes/useCommitVotes.ts
daywiss: cool good to know

PR #57, opened 2022-11-07
feat: add mobile styles

ryanwolhuter:
### Summary

Adds new mobile styles for the dapp.

Testing and reviewing this will be the subject of a whole round of QA that I will be co-ordinating, so don't worry about verifying the visual correctness of this while reviewing the code (you'd be doing double work).

I have been careful not to mess with any logic here, but I'll draw attention to a couple of things worth noting in comments in the PR.

### DIFF
--- a/graph/queries/getPastVotes.ts
+++ b/graph/queries/getPastVotes.ts
@@ -32,7 +32,7 @@ export async function getPastVotesV1() {
   return result?.priceRequests?.map(({ id, time, price, ancillaryData }) => {
     const identifier = getIdentifierFromPriceRequestId(id);
     const correctVote = Number(
-      formatVoteStringWithPrecision(parseEther(price), identifier)
+      formatVoteStringWithPrecision(parseEtherSafe(price), identifier)
     );
 
     return {


components/GlobalStyle/GlobalStyle.ts
```diff
@@ -244,10 +248,23 @@ a:not([class]) {
...
+    --desktop-page-width: ${desktopPageWidth}px;
+    --mobile-header-height: ${mobileHeaderHeight}px;
+    --desktop-header-height: ${desktopHeaderHeight}px;
+    --mobile-banner-height: ${mobileBannerHeight}px;
+    --desktop-banner-height: ${desktopBannerHeight}px;
+    --page-width: var(--desktop-page-width);
+    --header-height: var(--desktop-header-height);
+    --banner-height: var(--desktop-banner-height);
+    @media ${tabletAndUnder} {
+      --page-width: var(--mobile-page-width);
```
ryanwolhuter, line 261: we set the global container dimensions at the highest level so the rest of the components don't need to think about it

components/HowItWorks/HowItWorks.tsx
```diff
@@ -47,28 +48,22 @@ export function HowItWorks() {
...
+              </LoadingSkeleton>{" "}
               UMA tokens of{" "}
-              <Strong>
-                {isLoading() ? (
-                  <LoadingSkeleton width={60} />
-                ) : (
-                  formatNumberForDisplay(totalTokens())
-                )}
-              </Strong>{" "}
+              <LoadingSkeleton isLoading={isLoading()} width={60}>
```
ryanwolhuter, line 64: See the changes in `LoadingSkeleton` for details about these changes

components/LoadingSkeleton/LoadingSkeleton.tsx
```diff
@@ -1,30 +1,47 @@
...
 interface Props {
-  width?: string | number;
-  height?: string | number;
+  children: JSX.Element;
+  isLoading: boolean;
+  height?: CSSProperties["height"];
+  width?: CSSProperties["width"];
   variant?: "grey" | "white";
 }
 export function LoadingSkeleton({
```
ryanwolhuter, line 12: I reworked this so that you pass the content that it will be placeholder for as children. This lets us get the height of the loader from the child, and we don't have to specify different heights for different breakpoints. We can let the fluid typography handle it for us.

components/Panel/MenuPanel/MenuPanel.tsx
```diff
@@ -225,6 +240,18 @@ const Address = styled.p`
   font: var(--text-sm);
   margin-top: 10px;
   margin-bottom: 15px;
+
+  @media ${mobileAndUnder} {
+    display: none;
+  }
+`;
+
+const TruncatedAddress = styled(Address)`
```
ryanwolhuter, line 249: Since checking the screen width with the `useWindowSize` hook only works once the page is mounted, I prefer to use this pattern where possible. we have two components and we show / hide them on mobile with media queries and css (doesn't need the page to mount to work).

components/Tooltip/Tooltip.tsx
```diff
@@ -24,14 +24,30 @@ export function Tooltip({ children, ...rest }: Props) {
...
+            />
+          )
       )}
     </>
   );
 }
 
 const AnimatedTooltipContent = animated(TooltipPopup);
 
+const LabelWrapper = styled.div`
```
ryanwolhuter, line 42: Added this wrapper to the tooltip label so that we can do hyphenated word breaks — we need this for example when the vote is a super long number which has no word breaks

components/VotesList/VotesList.tsx
```diff
@@ -0,0 +1,37 @@
+import { tabletAndUnder, tabletMax } from "constant";
```
ryanwolhuter, line 1: This is the same `VotesTable` component as before. I renamed it because it is not always a table — on mobile it is a list of cards. The behavior is the same on desktop and mobile other than using different element types.

components/VotesList/VotesListItem.tsx
```diff
@@ -273,18 +293,57 @@ export function VotesTableRow({
   );
 }
 
+function VoteText({ voteText }: { voteText: string | undefined }) {
```
ryanwolhuter, line 296: If a vote is very long we truncate it and add a tooltip that shows the full value

constant/styles/fonts.ts
```diff
@@ -5,8 +5,8 @@ const bold = 700;
 
 const family = "Halyard Display";
 
-export const headerXl = `${bold} 80px/110px ${family}`;
-export const headerLg = `${bold} 40px/55px ${family}`;
+export const headerXl = `${bold} clamp(40px, 10vw, 80px)/clamp(55px, 8vw, 110px) ${family}`;
```
ryanwolhuter, line 8: This is fluid typography. We clamp the size of the headings so that they don't get too big or too small. They react to the screen size appropriately without needing any media queries.

daywiss (approved):
don't have much opinion on the code, but lgtm

PR #59, opened 2022-11-08
improve: handle malformed ancillary data

ryanwolhuter:
### Summary

Decoding hex strings can throw when the input data is malformed. Add try/catch logic that adds warnings in that case.


daywiss (approved):
nice

PR #60, opened 2022-11-08
feat: clear errors on success

ryanwolhuter:
### Summary

Add logic to clear errors from a particular origin when another transaction from that origin is successful.

### DIFF
--- a/components/Wallet/Wallet.tsx
+++ b/components/Wallet/Wallet.tsx
@@ -1,9 +1,10 @@
-import { useConnectWallet, useWallets } from "@web3-onboard/react";
-import { signingMessage } from "constant";
+import { useConnectWallet, useSetChain, useWallets } from "@web3-onboard/react";
+import { signingMessage, wrongChainMessage } from "constant";
 import { ethers } from "ethers";
 import { derivePrivateKey, recoverPublicKey } from "helpers";
 import {
   useContractsContext,
+  useErrorContext,
   usePanelContext,
   useUserContext,
   useWalletContext,
@@ -19,11 +20,27 @@ import { WalletIcon } from "./WalletIcon";
 
 export function Wallet() {
   const [{ wallet, connecting }, connect] = useConnectWallet();
+  const [{ connectedChain }, setChain] = useSetChain();
   const connectedWallets = useWallets();
   const { setProvider, setSigner, setSigningKeys } = useWalletContext();
   const { setVoting, setVotingToken } = useContractsContext();
   const { openPanel } = usePanelContext();
   const { address, truncatedAddress } = useUserContext();
+  const { addErrorMessage, removeErrorMessage } = useErrorContext();
+
+  useEffect(() => {
+    if (!connectedChain) return;
+
+    void (async () => {
+      if (connectedChain.id !== "0x5") {
+        addErrorMessage(wrongChainMessage);
+        await setChain({ chainId: "0x5" });
+      } else {
+        removeErrorMessage(wrongChainMessage);
+      }
+    })();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [connectedChain?.id, setChain]);
 
   useEffect(() => {
     if (!connectedWallets.length) return;
--- a/constant/index.ts
+++ b/constant/index.ts
@@ -1,71 +1,13 @@
-export { graphEndpoint } from "./query/graphEndpoint";
-export {
-  activeVotesKey,
-  balancesQueryKeys,
-  committedVotesForDelegatorKey,
-  committedVotesKey,
-  contentfulDataKey,
-  decryptedVotesKey,
-  delegateToStakerKey,
-  delegatorSetEventForDelegateKey,
-  delegatorSetEventsForDelegatorKey,
-  encryptedVotesKey,
-  hasActiveVotesKey,
-  ignoredRequestToBeDelegateAddressesKey,
-  outstandingRewardsKey,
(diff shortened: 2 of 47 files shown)


components/Wallet/Wallet.tsx
```diff
@@ -19,11 +20,27 @@ import { WalletIcon } from "./WalletIcon";
...
+  const [{ connectedChain }, setChain] = useSetChain();
   const connectedWallets = useWallets();
   const { setProvider, setSigner, setSigningKeys } = useWalletContext();
   const { setVoting, setVotingToken } = useContractsContext();
   const { openPanel } = usePanelContext();
   const { address, truncatedAddress } = useUserContext();
+  const { addErrorMessage, removeErrorMessage } = useErrorContext();
+
+  useEffect(() => {
+    if (!connectedChain) return;
```
ryanwolhuter, line 32: I added this to check that the logic for the generic data fetching error message works, and because people keep complaining about the errors when they connect to the wrong network lol

constant/index.ts
```diff
@@ -1,71 +1,13 @@
...
-  textXs,
-} from "./styles/fonts";
-export { shadow1, shadow2, shadow3 } from "./styles/shadows";
-export { discordLink } from "./voting/discordLink";
-export { earlyRequestMagicNumber } from "./voting/earlyRequestMagicNumber";
-export { numPhases, phaseLength, phaseLengthMilliseconds, roundLength } from "./voting/voteTiming";
-export { votingContractAddress, votingTokenContractAddress } from "./web3/addresses";
-export { goerliDeployBlock } from "./web3/deployBlocks";
-export { signingMessage } from "./web3/signingMessage";
+export * from "./misc/errorMessages";
```
ryanwolhuter, line 1: so much cleaner. the previous was autogenerated by a tool, but this is much better

contexts/ErrorContext.tsx
```diff
@@ -1,14 +1,25 @@
 import { createContext, ReactNode, useState } from "react";
+import { ErrorOriginT } from "types";
+
+type ErrorMessagesT = Record<ErrorOriginT, string[]>;
 
 export interface ErrorContextState {
```
ryanwolhuter, line 6: Using react nodes for error messages is problematic because then the messages can't be compared directly with each other (they are different objects). I changed to strings because we don't need to pass react nodes anywhere else in the code, and this makes things much simpler.

hooks/helpers/useHandleError.ts
```diff
@@ -1,16 +1,48 @@
...
+    errorOrigin,
+    isContractTransaction = true,
+    isDataFetching = false,
+    customErrorMessage,
+  } = options || {};
+  const { addErrorMessage, clearErrorMessages, errorMessages } =
+    useErrorContext(errorOrigin);
+
+  function onError(error: unknown) {
+    if (errorMessages.includes(wrongChainMessage)) return;
```
ryanwolhuter, line 23: If the user is on the wrong chain, we might as well not show any other errors. Nothing will work until they get onto the correct chain.

hooks/helpers/useHandleError.ts
```diff
@@ -1,16 +1,48 @@
...
+
+    let message: string;
 
-  return (error: unknown) => {
-    if (error instanceof Error) {
-      addErrorMessage(error.message);
+    if (isDataFetching) {
+      message = errorFetchingVoteDataMessage;
```
ryanwolhuter, line 28: Instead of showing a different error for every data fetching error that happens, show one generic error message. Since data fetching either works or it doesn't (we need to load all the vote data from the chain for the dapp to work) there's no point in showing 10 different errors that all say some variation of "error fetching x".

hooks/helpers/useHandleError.ts
```diff
@@ -1,16 +1,48 @@
...
+    } else if (customErrorMessage) {
+      message = customErrorMessage;
+    } else if (error instanceof Error) {
+      message = isContractTransaction
```
ryanwolhuter, line 32: I moved the error formatting logic here for convenience

hooks/mutations/delegation/useAcceptReceivedRequestToBeDelegate.ts
```diff
@@ -5,16 +5,19 @@ import { setDelegator } from "web3";
 
 export function useAcceptReceivedRequestToBeDelegate() {
   const { address } = useUserContext();
-  const onError = useHandleError();
+  const { onError, clearErrors } = useHandleError();
   const queryClient = useQueryClient();
 
   const { mutate, isLoading } = useMutation(setDelegator, {
     onError,
     onSuccess: (_data, { delegatorAddress }) => {
```
ryanwolhuter, line 13: This pattern is added to all of the mutations — when the transaction succeeds, clear the errors for that context

hooks/queries/delegation/useCommittedVotesForDelegator.ts
```diff
@@ -14,7 +14,7 @@ export function useCommittedVotesForDelegator() {
   const { voting } = useContractsContext();
   const { getDelegationStatus, getDelegatorAddress } = useDelegationContext();
   const { roundId } = useVoteTimingContext();
-  const onError = useHandleError();
+  const { onError } = useHandleError({ isDataFetching: true });
```
ryanwolhuter, line 17: For queries, we pass `isDataFetching: true` so that the handler knows to show the generic data fetching error message.

daywiss (approved):

PR #63, opened 2022-11-09
chore: upgrade to nextjs 13

ryanwolhuter:
### Summary

Bumps react, react-dom and nextjs https://nextjs.org/blog/next-13

Updates usages of "next/link" to use the new improved pattern which doesn't need an `a` tag as a child and doesn't need to have `passHref`.


daywiss (approved):
not too bad

PR #65, opened 2022-11-10
feat: add polymarket options

ryanwolhuter:
### Summary

When a yes or no query comes from polymarket, it should have a standard set of options which are included in the ancillary data. I have identified two types which are identical except that one of them also includes an option p4 which corresponds to the early request magic number.

We can use this information to create and show a dropdown with options instead of a simple text input.

* Add a function that checks if the ancillary data for a polymarket request conforms to one of these known option types and generates options for a dropdown.


components/VotesList/VotesListItem.tsx
```diff
@@ -63,9 +64,12 @@ export function VotesListItem({
   const maxDecimals = getPrecisionForIdentifier(decodedIdentifier);
   const Icon = origin === "UMA" ? UMAIcon : PolymarketIcon;
 
-  function formatTitle(title: string) {
-    if (title.length <= 45) return title;
-    return `${title.substring(0, 45)}...`;
+  function onSelectVote(option: DropdownItemT) {
```
ryanwolhuter, line 67: If the user chooses to do a custom input, then change the input from the dropdown to a normal text input.

helpers/voting/formatVotes.ts
```diff
@@ -121,5 +121,7 @@ export function formatVoteStringWithPrecision(
   const identifierPrecision = BigNumber.from(
     getPrecisionForIdentifier(decodedIdentifier)
   ).toString();
-  return formatFixed(vote, identifierPrecision).toString();
+  const formatted = formatFixed(vote, identifierPrecision).toString();
+  // if the formatted number ends with .0, remove the .0
```
ryanwolhuter, line 125: Ethers formats whole numbers with decimal points like "1.0" which breaks the matching logic. Remove that redundant decimal if present.

PR #67, opened 2022-11-14
improve: prevent users with 0 stake from voting

ryanwolhuter:
### Summary

If a user who has not staked UMA votes, they will have no impact on the outcome. Disable the commit / reveal button in this case and add a tooltip explaining the situation.

* Added a tooltip that explains why the commit / reveal button is disabled
* Added explanations for other situations where the button is disabled as well.


constant/query/queryKeys.ts
```diff
@@ -19,13 +19,17 @@ export const unstakeCoolDownKey = "unstakeCoolDownKey";
 // user
 export const userDataKey = "userDataKey";
 // delegation
-export const delegatorSetEventForDelegateKey = "delegatorSetEventForDelegateKey";
-export const delegatorSetEventsForDelegatorKey = "delegatorSetEventsForDelegatorKey";
-export const receivedRequestsToBeDelegateKey = "receivedRequestsToBeDelegateKey";
+export const delegatorSetEventForDelegateKey =
```
ryanwolhuter, line 22: this file is unchanged except for applied formatting

md0x (approved):
Looks good!

PR #69, opened 2022-11-15
bug: maximum approve shows huge number that overflows input

ryanwolhuter:
### Summary

When clicking "max" in the approve input, we use the largest allowed uint256 as the value (essentially the same as approving an unlimited amount). However, this shows a very large number in the input which is ugly and confusing to the user.

* Add a string value for "unlimited approval" and show that in the input instead
* Handle this string as a possible input for the approve mutation

### DIFF
--- a/components/Input/AmountInput.tsx
+++ b/components/Input/AmountInput.tsx
@@ -1,7 +1,7 @@
 import { Button } from "components";
 import UmaToken from "public/assets/icons/uma-token.svg";
 import styled from "styled-components";
-import { useOnChange } from "hooks";
+import { useHandleDecimalInput } from "hooks";
 import { Input, Wrapper } from "./Input";
 
 interface Props {
@@ -22,7 +22,7 @@ export function AmountInput({
   allowNegative = true,
   maxDecimals = 18,
 }: Props) {
-  const onChange = useOnChange(onInput, maxDecimals, allowNegative);
+  const onChange = useHandleDecimalInput(onInput, maxDecimals, allowNegative);
 
   return (
     <_Wrapper aria-disabled={disabled}>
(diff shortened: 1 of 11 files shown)


components/Input/AmountInput.tsx
```diff
@@ -1,7 +1,7 @@
 import { Button } from "components";
 import UmaToken from "public/assets/icons/uma-token.svg";
 import styled from "styled-components";
-import { useOnChange } from "hooks";
+import { useHandleDecimalInput } from "hooks";
```
ryanwolhuter, line 4: Renamed this for clarity. David suggested this ages ago and it somehow slipped through the cracks.

components/Panel/StakeUnstakePanel/StakeUnstakePanel.tsx
```diff
@@ -53,7 +55,10 @@ export function StakeUnstakePanel() {
   }
 
   function approve(approveAmountInput: string) {
-    const approveAmount = parseEtherSafe(approveAmountInput);
+    const approveAmount =
```
ryanwolhuter, line 58: We recognize the maximum approval amount string and react accordingly

hooks/helpers/useHandleDecimalInput.ts
```diff
@@ -10,13 +11,17 @@ export function useOnChange(
...
   return (event: ChangeEvent<HTMLInputElement>) => {
-    const value = event.target.value;
+    let value = event.target.value;
 
     if (type !== "number") {
       onInput(value);
       return;
     }
 
+    if (value.includes(maximumApprovalAmountString)) {
```
ryanwolhuter, line 21: If the input has the maximum approval amount string in it because that was what was input previously, it will look like "Unlimited approval2" etc. We remove the unlimited approval string and use just the number. This effectively clears the unlimited approval input if the user decides to type a different number instead.

web3/mutations/staking/approve.ts
```diff
@@ -14,9 +18,14 @@ export async function approve({
     votingContractAddress,
     approveAmount
   );
+
+  const amountToDisplay = approveAmount.eq(maximumApprovalAmount)
```
ryanwolhuter, line 22: Show "unlimited" instead of the super long string in the notification

md0x (approved):
LGTM

PR #71, opened 2022-11-16
feat: add polling

ryanwolhuter:
### Summary

Luckily for us and our infura quota, there are very few places where polling the chain is necessary. In general we fetch data on load, when reconnecting, or when navigating back to the page. Since react query lets us use whatever values as query keys, we can use the `address` as a key which triggers a refetch when changed. Same goes for the `roundId`, which means that the dapp already knows to refetch vote data when the round moves.

For monitoring newly added upcoming votes, we subscribe to the `PriceRequestAdded` event in ethers. We start up the dapp with a counter set to 0, and then increment that counter whenever a new request is added. We use this number as a query key for the `useUpcomingVotes` hook, which refetches when the number changes. We also use this pattern for monitoring new received requests to be delegate.

The only polling that happens more frequently than once a minute on the chain happens in the query for active votes. Since there is a delay between when the round advances and getPendingRequests returning the new active requests, we need to refetch until it does. I have tried this with smaller numbers and I found that 30 seconds was the lowest I could go while still having it work consistently.

**NOTE:** this polling will ONLY run in the event that the user was sitting with the dapp open watching the round roll into the new round. So we're not doing this every time the round starts for all users, just for the rare case that someone was staring at the dapp at this exact 30 second long moment.


hooks/queries/rewards/useRewardsCalculationInputs.ts
```diff
@@ -0,0 +1,42 @@
...
+import { BigNumber } from "ethers";
+import {
+  useContractsContext,
+  useHandleError,
+  useStakingContext,
+  useUserContext,
+} from "hooks";
+import { getRewardsCalculationInputs } from "web3";
+
+export function useRewardsCalculationInputs() {
```
ryanwolhuter, line 12: query that returns the data we need as initial inputs for calculating rewards

hooks/queries/user/useUserVotingAndStakingDetails.ts
```diff
@@ -13,7 +13,7 @@ export function useUserVotingAndStakingDetails() {
     () => getUserData(address),
     {
       enabled: !!address,
-      refetchInterval: (data) => (data ? false : 100),
+      refetchInterval: oneMinute,
```
ryanwolhuter, line 16: since hitting the graph is cheap, we might as well poll this in case its changed from actions taken elsewhere

hooks/queries/votes/usePastVotes.ts
```diff
@@ -11,7 +11,7 @@ export function usePastVotes() {
     [pastVotesKey, roundId],
     () => getPastVotesAllVersions(),
     {
-      refetchInterval: (data) => (data ? false : 100),
+      refetchInterval: oneMinute,
```
ryanwolhuter, line 14: hitting the graph is cheap so might as well poll past votes (there is a delay between a round ending and the events being indexed by the graph)

chrismaree (approved):
this is extremely cool! I really like being able to see the rewards tick up as you look at the dapp. also, polling inbound requests the way you have here is super cool.

PR #73, opened 2022-11-17
feat: update request augmentation and add cache warming endpoint

chrismaree:
This PR adds a serverless function endpoint used to warm the other heavy endpoints to speed up the cache of heavy operations. Vercel will cache the responces for a given request for 30 days. therefor we simply need to run this endpoint once after deployment and the cache will be built and be fast!


ryanwolhuter (approved):
cool and nice

PR #76, opened 2022-11-17
improve: add date and time to votes

ryanwolhuter:
### Summary

Include a formatted date and time in each of the votes in the votes list.

Screenshot 2022-11-17 at 14 51 08


chrismaree (approved):
looks good! it might be useful to get some input on design for this. there are additional improvements we can make but dont need to worry about that rn.

PR #77, opened 2022-11-17
bug: should show across v2 as identifier

ryanwolhuter:
### Summary

For reasons I cannot remember, I decided to use the internal graphql id for votes to get the identifiers of past votes. This makes no sense because we literally already get the actual value from the graph, and also because it introduced a bug where we split the internal identifier string which doesn't work consistently.

* Remove the query for the internal ID
* Use the identifier returned from the graph as-is

### DIFF
--- a/types/queries.ts
+++ b/types/queries.ts
@@ -1,6 +1,5 @@
 export type PastVotesQuery = {
   priceRequests: {
-    id: string;
     identifier: {
       id: string;
     };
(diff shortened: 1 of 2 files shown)


types/queries.ts
```diff
@@ -1,6 +1,5 @@
 export type PastVotesQuery = {
   priceRequests: {
-    id: string;
```
chrismaree, line 3: nice. smaller payload more better

chrismaree (approved):

chrismaree:
looks good! tested on staging link and identifiers are correct now.

PR #78, opened 2022-11-18
feat: use decode admin serverless function

ryanwolhuter:
### Summary

Use the /decode-admin serverless function to get the transactions associated with admin requests.

* Add getter and hook for fetching this data (non-blocking)
* Display this data in the vote details panel if and when it exists

Screenshot 2022-11-18 at 09 43 15


components/Panel/VotePanel/Details.tsx
```diff
@@ -49,15 +52,28 @@ export function Details({
...
+          <PanelSectionTitle>
+            <IconWrapper>
+              <VoteIcon />
+            </IconWrapper>{" "}
+            Decoded ancillary data
+          </PanelSectionTitle>
+          <Text> {decodedAncillaryData} </Text>
+        </SectionWrapper>
+      )}
+      {decodedAdminTransaction && (
```
ryanwolhuter, line 66: show decoded admin transaction as pre-formatted text if it exists

components/Panel/VotePanel/VotePanel.tsx
```diff
@@ -36,16 +30,7 @@ export function VotePanel({ content }: Props) {
...
-        <Details
-          description={description}
-          decodedAncillaryData={decodedAncillaryData}
-          options={options}
-          timeAsDate={timeAsDate}
-          links={links}
-          discordLink={discordLink}
-        />
-      ),
+      content: <Details {...content} />,
```
ryanwolhuter, line 33: using spread syntax for convenience because there are a lot of props

constant/query/queryKeys.ts
```diff
@@ -31,31 +32,3 @@ export const delegateToStakerKey = "delegateToStakerKey";
 export const ignoredRequestToBeDelegateAddressesKey =
   "ignoredRequestToBeDelegateAddressesKey";
 export const committedVotesForDelegatorKey = "committedVotesForDelegatorKey";
-
```
ryanwolhuter, line 34: Removed these because we don't actually use them anywhere

chrismaree:
I tried out this PR on the staging link but it does not seem to work. I got the following errors:
2022-11-18 at 12 30 53@2x
2022-11-18 at 12 30 50@2x

I'm investigating as to why this happened.

chrismaree:
> I tried out this PR on the staging link but it does not seem to work. I got the following errors: 2022-11-18 at 12 30 53@2x 2022-11-18 at 12 30 50@2x
> 
> I'm investigating as to why this happened.

it looks like we are missing some environment variables in the vercel deployment :) 
2022-11-18 at 12 32 16@2x

chrismaree:
after playing around with this on the staging link I think things are looking correct now (the right data is within each admin call) but I have some feedback on how we are displaying this info:

1. You are missing critical information returned in the decoded transactions that we should be showing. the Decoded tx call includes `data, to, value and decoded data`. you are currently only showing `decodedData`. We should show the `to` prop as well, at minimum. This can be combined with the subsequent suggestion.
2. Often, an admin action will contain multiple transactions. the sereverless function that you are using returns an array of transactions which you've concatenated into one big block. I think it would be more illustrative to the contents of the transaction if you did not concatenate these and rather spread them with a bit more meta data. this combinds with feedback point 1 on needing to include the `to` prop. 


let me show you what I'm thinking with a screenshot. below is what we currently have for admin `176`
2022-11-22 at 10 56 53@2x


What I think we should do is as follows(recommendations in red):
2022-11-22 at 10 59 10@2x


The key differences here are:
1) you show the index in the array for the given transaction (rather than loosing it by contamination between the blocks)
2) you MUST show the `to` prop. this is really important. Ideally this is made as a hyperlink to etherscan


Note that we might want to remove some of the formatting that comes back from the serverless function (such as `Nested transactions`) and rather make this part of the rendered UI component. Curious what you think on this.

chrismaree:
a second bit of feedback is that we are actually loosing some critical UI data in how we are currently displaying these values: the Identifier and ancillary data (this is actually a comment thing I've realized across all requests). we _should_ always show an identifier and ancillary data field in the sidebar. For example, on the admin identifiers I cant actually see what the identifier and ancillary data is. the identifier is the something like `ADMIN-x` and show _always_ be shown somewhere in the side bar. the new governor contracts in DVM2.0 includes ancillary data which should _also_ be shown.

chrismaree:
one final bit of feedback: I think that we should consider using an accordion component or some kind of fold to hide the tx data unless the user chooses to expand it. in the current structure it is hard to see what is going on with big transactions. ideally, this is collapsed by default and then you can click on the tx data to expand it.

a follow on bit of feedback is that the current design starts scrolled all the way to the bottom of the panel when you open admin votes with a lot of detail. it should start scrolled to the top, not the bottom.

chrismaree:
Regarding the 3 comments I left above I'll push some changes now to implement this. I'd like to get my hands a bit dirty with the front end code :)

chrismaree:
I implemented one of the comments listed above (numbering the transactions and including the `to` address). The other three (folding block, scrolling and including the admin identifier & ancillary) I did not add. decoded transactions now look like this:
2022-11-22 at 11 40 06

Note that my code is bad and should likely be improved :) I was not sure if you already had an address shortening function or an etherscan link building function so I hackily added these to show what I want it to look like. please improve.

chrismaree (approved):

web3/queries/votes/getDecodedAdminTransactions.ts
```diff
@@ -0,0 +1,32 @@
...
+      "Content-Type": "application/json",
+    },
+    body: JSON.stringify({ identifiers: decodedIdentifiers }),
+  });
+
+  if (!response.ok) {
+    throw new Error("Error fetching decoded admin transactions");
+  }
+
+  const result = (await response.json()) as DecodedAdminTransactionsT[];
```
daywiss, line 21: if we have a shape for this data we could use a validator to enforce this, probably a todo item

daywiss (approved):

PR #79, opened 2022-11-18
improve: make vote table columns narrower

ryanwolhuter:
### Summary

I've added some fancy css that makes the page and the vote table smarter about how it handles spacing. Now it determines how much space items should take in the table and how much padding the page should have based on a combination of our preferred values from the design, the width of the page, and the smallest value that doesn't break the design.

### DIFF
--- a/components/Banner/Banner.tsx
+++ b/components/Banner/Banner.tsx
@@ -1,4 +1,4 @@
-import { mobileAndUnder } from "constant";
+import { tabletAndUnder } from "constant";
 import { ReactNode } from "react";
 import styled from "styled-components";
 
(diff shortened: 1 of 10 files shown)


components/Banner/Banner.tsx
```diff
@@ -24,10 +24,9 @@ const InnerWrapper = styled.div`
   height: var(--banner-height);
   display: flex;
   align-items: center;
-  padding-left: 45px;
-  @media ${mobileAndUnder} {
-    max-width: unset;
-    padding: 15px;
+  padding-left: clamp(10px, 45px, 4vw);
```
ryanwolhuter, line 27: One of my favorite css magic tricks https://developer.mozilla.org/en-US/docs/Web/CSS/clamp

components/VotesList/VotesListItem.tsx
```diff
@@ -178,6 +178,17 @@ export function VotesListItem({
     return "transparent";
   }
 
+  function getTitleMaxWidth() {
```
ryanwolhuter, line 181: different max width algorithms depending on how much space is available in the table

stories/Pages/VotePage/VotePage.stories.tsx
```diff
@@ -1,11 +1,99 @@
-import { ComponentMeta, ComponentStory } from "@storybook/react";
+import { Meta, Story } from "@storybook/react";
```
ryanwolhuter, line 1: storybook mocks, you can ignore this

PR #80, opened 2022-11-18
feat: add metadata for across requests

ryanwolhuter:
### Summary

Show customized data when we see the "ACROSS-V2" identifier.

* Provide a yes/no dropdown for vote options
* Show the across icon in the vote
* Use the description from the docs and provide a learn more link
* Add a link to umip 157


helpers/voting/getVoteMetaData.ts
```diff
@@ -44,6 +44,30 @@ export function getVoteMetaData(
     };
   }
 
+  const isAcross = decodedIdentifier === "ACROSS-V2";
+  if (isAcross) {
```
chrismaree, line 48: This is fine for now but it might make sense to have some external module where we put these kinds of encodings/ augmentations so that it does not live within the `getVoteMetaData file.

chrismaree (approved):

PR #91, opened 2022-11-22
feat: add past rewards retrival serverless function to build calldata

chrismaree:
This PR adds a serverless function to build call data to fetch voting rewards from previously deployed contracts, used by the front end.


ryanwolhuter (approved):
looks good, lets merge this now and I'll do the implementation in a separate PR

PR #93, opened 2022-11-23
fix: update reward retrival with new logic

chrismaree:
This PR makes a small modification to the past rewards retrieval serverless function after testing it with the goerli deployments.


ryanwolhuter (approved):
lgtm

PR #98, opened 2022-11-24
fix: unstaking max button sets an amount higher than staked balance

ryanwolhuter:
### Summary

There was some trouble with the max button for staking / unstaking. Fix these issues by getting the staked balance by calling `getVoterStakePostUpdate()` to make sure its accurate, and also add some guard logic to make sure that we never attempt to stake / unstake more than the staked / unstaked balance when using the max button.

### DIFF
--- a/components/Panel/StakeUnstakePanel/Stake.tsx
+++ b/components/Panel/StakeUnstakePanel/Stake.tsx
@@ -15,7 +15,7 @@ import {
 interface Props {
   tokenAllowance: BigNumber | undefined;
   unstakedBalance: BigNumber | undefined;
-  unstakeCoolDown: number | undefined;
+  unstakeCoolDown: BigNumber | undefined;
   isDelegate: boolean;
   approve: (approveAmount: string) => void;
   stake: (stakeAmount: string, resetStakeAmount: () => void) => void;
@@ -31,7 +31,7 @@ export function Stake({
   const [inputAmount, setInputAmount] = useState("");
   const [disclaimerChecked, setDisclaimerChecked] = useState(false);
   const unstakeCoolDownFormatted = unstakeCoolDown
-    ? formatDuration({ seconds: unstakeCoolDown })
+    ? formatDuration({ seconds: unstakeCoolDown.toNumber() })
     : "0 seconds";
 
   const disclaimer = `I understand that Staked tokens cannot be transferred for ${unstakeCoolDownFormatted} after unstaking.`;
--- a/components/Panel/StakeUnstakePanel/StakeUnstakePanel.tsx
+++ b/components/Panel/StakeUnstakePanel/StakeUnstakePanel.tsx
@@ -1,5 +1,4 @@
-import { Tabs } from "components";
-import { LoadingSkeleton } from "components";
+import { LoadingSkeleton, Tabs } from "components";
 import { maximumApprovalAmountString } from "constant";
 import { formatNumberForDisplay, parseEtherSafe } from "helpers";
 import { maximumApprovalAmount } from "helpers/web3/ethers";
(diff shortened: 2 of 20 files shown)


components/Panel/StakeUnstakePanel/Stake.tsx
```diff
@@ -15,7 +15,7 @@ import {
 interface Props {
   tokenAllowance: BigNumber | undefined;
   unstakedBalance: BigNumber | undefined;
-  unstakeCoolDown: number | undefined;
+  unstakeCoolDown: BigNumber | undefined;
```
ryanwolhuter, line 18: updated these queries to use BigNumber for consistency

components/Panel/StakeUnstakePanel/StakeUnstakePanel.tsx
```diff
@@ -64,7 +63,13 @@ export function StakeUnstakePanel() {
   }
 
   function stake(stakeAmountInput: string, resetStakeAmount: () => void) {
-    const stakeAmount = parseEtherSafe(stakeAmountInput);
+    if (!unstakedBalance) return;
+    const parsedStakeAmountInput = parseEtherSafe(stakeAmountInput);
```
ryanwolhuter, line 67: handle potentially too large number input when using max

constant/query/queryKeys.ts
```diff
@@ -1,9 +1,7 @@
 // voting - not user dependent
-export const hasActiveVotesKey = "hasActiveVotesKey";
```
ryanwolhuter, line 2: removed some unused keys

contexts/StakingContext.tsx
```diff
@@ -19,7 +20,8 @@ export interface StakingContextState {
   tokenAllowance: BigNumber | undefined;
   unstakeRequestTime: Date | undefined;
   canUnstakeTime: Date | undefined;
-  unstakeCoolDown: number | undefined;
+  unstakeCoolDown: BigNumber | undefined;
+  resetOutstandingRewards: () => void;
```
ryanwolhuter, line 24: I realized that I did not update the query updater logic to reset the outstanding rewards that we are calculating instead of querying

hooks/mutations/staking/useExecuteUnstake.ts
```diff
@@ -17,47 +21,46 @@ export function useExecuteUnstake(errorOrigin?: ErrorOriginT) {
 
   const { mutate, isLoading } = useMutation(executeUnstake, {
     onError,
-    onSuccess: () => {
+    onSuccess: (_contractReceipt, { pendingUnstake }) => {
```
ryanwolhuter, line 24: was able to simplify this logic a bit

hooks/mutations/staking/useStake.ts
```diff
@@ -15,18 +15,14 @@ export function useStake(errorOrigin?: ErrorOriginT) {
     onSuccess: (_data, { stakeAmount }) => {
       clearErrors();
 
-      queryClient.setQueryData<StakerDetailsT>(
-        [stakerDetailsKey, address],
-        (oldStakerDetails) => {
-          if (oldStakerDetails === undefined) return;
+      queryClient.setQueryData<BigNumber>(
```
ryanwolhuter, line 18: need to update staked balance query directly now in mutations

components/Panel/StakeUnstakePanel/StakeUnstakePanel.tsx
md0x: I'm trying to find out if there might not be a deeper problem that makes `parsedStakeAmountInput.gt(unstakedBalance)===true`

Do you think this treatment could be affecting the accuracy of the numbers?
[parseEtherSafe](https://github.com/UMAprotocol/voter-dapp-v2/blob/37caf4e710238a123ea54450fb087d85c4cd300a/helpers/web3/ethers.ts#L11)

Have you tried using here, instead of `parseEtherSafe`, ethers like:
`const parsedStakeAmountInput = ethers.utils.parseEther(stakeAmountInput);`

ryanwolhuter: Yeah that could be, since there's some casting happening there. I'll try this and let you know

ryanwolhuter: You were right 👍

md0x (approved):
Lgtm!

PR #99, opened 2022-11-25
feat: add total rewards fetched

chrismaree:
This PR adds the ability for the `past-rewards` serverless function to include the total amount of rewards fetched. It also includes a small formatting improvement.


pages/api/past-rewards.ts
```diff
@@ -75,7 +74,16 @@ async function constructMultiCall(
       ])
     );
   });
-  return multiCallPayload;
+
+  try {
```
chrismaree, line 78: here is the primary change of the PR.

ryanwolhuter (approved):

PR #100, opened 2022-11-28
fix: panel scroll issue is back

ryanwolhuter:
### Summary

We want the side panel to always open with its content scrolled all the way to the top. This proved to be impossible with the [reach ui dialog library](https://reach.tech/dialog/) we were using because that library destroys and recreates the dom node of the panel's content. this means that we cannot access a ref to the content to perform a scroll action until _after_ the component is fully rendered, causing the scroll to happen after the panel slides in which looks ugly.

I've been getting quite frustrated with the inflexibility of the reach ui component for a while now. So I've decided to remove that lib and replace it with my own implementation. It turns out that the reach component doesn't actually do all that much, and I was able to address all of the [accessibility concerns](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role) in my custom implementation without much difficulty.

* replace reach ui dialog with a custom implementation

### DIFF
--- a/components/Panel/Panel.tsx
+++ b/components/Panel/Panel.tsx
@@ -1,4 +1,3 @@
-import "@reach/dialog/styles.css";
 import { black, white } from "constant";
 import { usePanelContext, usePanelWidth } from "hooks";
 import Close from "public/assets/icons/close.svg";
--- a/components/Panel/Panel.tsx
+++ b/components/Panel/Panel.tsx
@@ -1,9 +1,9 @@
-import { DialogContent, DialogOverlay } from "@reach/dialog";
 import "@reach/dialog/styles.css";
 import { black, white } from "constant";
 import { usePanelContext, usePanelWidth } from "hooks";
 import Close from "public/assets/icons/close.svg";
 import { CSSProperties, useEffect, useRef } from "react";
+import { FocusOn } from "react-focus-on";
 import { animated, useTransition } from "react-spring";
 import styled from "styled-components";
 import { ClaimPanel } from "./ClaimPanel";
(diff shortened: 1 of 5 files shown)


components/Panel/Panel.tsx
```diff
@@ -29,20 +28,18 @@ export function Panel() {
   const panelWidth = usePanelWidth();
   const contentRef = useRef<HTMLDivElement | null>(null);
 
+  const transitions = useTransition(panelOpen, {
```
ryanwolhuter, line 31: we're still using react spring's transition helper for animating mount and unmount of the panel's overlay as before.

components/Panel/Panel.tsx
```diff
@@ -51,57 +48,72 @@ export function Panel() {
...
-                  <IconWrapper>
-                    <CloseIcon />
-                  </IconWrapper>
-                </CloseButton>
-              </Content>
-            </Overlay>
+            ></Overlay>
           )
       )}
+      <FocusOn
```
ryanwolhuter, line 63: This component addresses the main accessibility concerns of a dialog like this by focusing the content and locking scroll of the document body.

contexts/PanelContext.tsx
```diff
@@ -10,7 +10,7 @@ export interface PanelContextState {
 }
 
 export const defaultPanelContextState = {
-  panelType: null,
+  panelType: "menu" as const,
```
ryanwolhuter, line 13: having a default panel type instead of null makes life easier (don't have to manage null values)

daywiss (approved):
very cool, works well as far as i can tell

PR #103, opened 2022-11-28
improve: add error handler for transaction

ryanwolhuter:
### Summary

When an error message comes from an ethers contract interaction with a [known error code](https://docs.ethers.io/v5/troubleshooting/errors/), we can parse the message to get the link to the error description in the ethers docs and build a clickable link from it.

* Add list of ethers error codes
* Add component that maybe renders a clickable link to the ethers docs if possible


daywiss (approved):
nice

PR #111, opened 2022-11-30
fix: past votes shows me that i did not vote

ryanwolhuter:
### Summary

There were some logical issues with how we fetched encrypted votes from the chain. Firstly, we were not using a checksum address, which caused ethers to think the address was incorrect (I could swear this worked before though so this is quite strange). Secondly, we were only fetching encrypted votes for the current round, which meant that we would not have encrypted votes for past votes. I fixed these problems, and also added another event query to fetch the encrypted votes from v1 as well — now you will see your vote in the past votes from v1 too.

* use checksum address in user context to prevent missed event queries
* get encrypted votes for all rounds to show in past votes
* get encrypted votes from v1 in the same query

### DIFF
--- a/web3/queries/votes/getEncryptedVotes.ts
+++ b/web3/queries/votes/getEncryptedVotes.ts
@@ -1,25 +1,31 @@
-import { VotingV2Ethers } from "@uma/contracts-frontend";
+import { VotingEthers, VotingV2Ethers } from "@uma/contracts-frontend";
 import { decodeHexString, makeUniqueKeyForVote } from "helpers";
 import { EncryptedVotesByKeyT } from "types";
 
 export async function getEncryptedVotes(
   votingContract: VotingV2Ethers,
-  address: string,
-  roundId: number
+  votingV1Contract: VotingEthers,
+  address: string
 ) {
-  const filter = votingContract.filters.EncryptedVote(
-    address,
-    roundId,
-    null,
-    null,
-    null,
-    null
-  );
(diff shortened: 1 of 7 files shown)


chrismaree (approved):
tested with the same account I used to create the bug and looks good now.

web3/queries/votes/getEncryptedVotes.ts
```diff
@@ -1,25 +1,31 @@
...
-  const eventData = result?.map(({ args }) => args);
   const encryptedVotes: EncryptedVotesByKeyT = {};
-  eventData?.forEach(({ encryptedVote, identifier, time, ancillaryData }) => {
+
+  v1EventData?.forEach(({ encryptedVote, identifier, time, ancillaryData }) => {
+    const decodedIdentifier = decodeHexString(identifier);
+    encryptedVotes[
+      makeUniqueKeyForVote(decodedIdentifier, time, ancillaryData)
+    ] = encryptedVote;
+  });
```
daywiss, line 26: my fear with this is that v1 vote data is somehow different of off from v2 data, which may require extra knowledge that its data coming from v1, but i guess we can cross that bridge if its a problem

daywiss (approved):

web3/queries/votes/getEncryptedVotes.ts
ryanwolhuter: yeah I agree, we'll have to keep an eye out. still, this is just for historic data so its not the end of the world if we encounter a bug here

PR #113, opened 2022-11-30
fix: disable claiming with amounts below threshold

daywiss:

Claim button was not being disabled when there wasnt anything to claim. This was tricky because technically there is a non 0 amount to claim, but it shows up in the UI as 0 because of truncation. We set a minimum value as a threshold of ".01" to claiming to make UI consistent with button behavior.


ryanwolhuter (approved):
looks good, I agree with your reasoning. Claiming less than this amount would make no sense anyway because the value would be less then the gas cost.

PR #115, opened 2022-12-01
fix: add even fancier table layout algorithm

ryanwolhuter:
The spacing in the table is now responsive based on the table's container size as well as the page size and page padding. We use css variables to set these and update them dynamically as needed. The vote title is now always as long as possible while still keeping the table width consistent.

### DIFF
--- a/components/VotesList/VotesListItem.tsx
+++ b/components/VotesList/VotesListItem.tsx
@@ -19,8 +19,8 @@ import Polymarket from "public/assets/icons/polymarket.svg";
 import Rolled from "public/assets/icons/rolled.svg";
 import UMAGovernance from "public/assets/icons/uma-governance.svg";
 import UMA from "public/assets/icons/uma.svg";
-import { ReactNode, useEffect, useState } from "react";
-import styled, { CSSProperties } from "styled-components";
+import { CSSProperties, ReactNode, useEffect, useRef, useState } from "react";
+import styled from "styled-components";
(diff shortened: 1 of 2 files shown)


components/VotesList/VotesListItem.tsx
```diff
@@ -191,15 +199,42 @@ export function VotesListItem({
...
-  function getTitleMaxWidth() {
-    if (activityStatus === "upcoming") return "70vw";
-    if (activityStatus === "active" && phase === "commit")
-      return "max(35vw, 320px)";
-    if (
-      (activityStatus === "active" && phase === "reveal") ||
-      activityStatus === "past"
-    )
-      return "max(400px, 45vw)";
+  function getCellWidths() {
```
ryanwolhuter, line 202: we now use different table cell widths depending on the activity status and phase. (used to only change the title's width)

components/VotesList/VotesListItem.tsx
```diff
@@ -221,29 +256,27 @@ export function VotesListItem({
...
-              "--border-color": getBorderColor(),
-            } as CSSProperties
-          }
-        >
+    <Wrapper
+      as={isTabletAndUnder ? "div" : "tr"}
+      style={style}
+      ref={wrapperRef}
+    >
+      <VoteTitleCell as={isTabletAndUnder ? "div" : "td"}>
```
ryanwolhuter, line 273: renamed `td` elements to `-Cell` for clarity

daywiss (approved):
works really well in my testing

PR #116, opened 2022-12-01
fix: filter v1 votes from history panel

daywiss:

This basically just filters out v1 votes from history panel. 

image


ryanwolhuter (approved):
💯

PR #121, opened 2022-12-02
fix: paginating one page also paginates the other pages

ryanwolhuter:
[Updating nested objects in react state](https://beta.reactjs.org/learn/updating-objects-in-state) is ugly. Trying to maintain an object of objects for the various pagination states was a mistake — it didn't work properly to begin with, and to make it work, we'd have to use multiple layers of object spreads all over the place.

Much easier and cleaner to just use separate state for each object.

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -4,7 +4,18 @@ import React from "react";
 import "styles/fonts.css";
 import { GlobalStyle } from "../components/GlobalStyle";
 import { Panel } from "../components/Panel/Panel";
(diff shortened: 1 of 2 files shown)


.storybook/preview.tsx
```diff
@@ -20,27 +31,28 @@ const queryClient = new QueryClient();
...
+      <WalletProvider>
+        <QueryClientProvider client={queryClient}>
+          <UserProvider>
+            <ContractsProvider>
+              <StakingProvider>
+                <PaginationProvider>
                   <DelegationProvider>
                     <VotesProvider>
-                    <PanelProvider>
-                      <PaginationProvider />
```
ryanwolhuter, line 32: this was a bug, no idea how or why it happened

PR #122, opened 2022-12-02
improve: remove error handlers from non-critical data fetchers

ryanwolhuter:
The app still works fine without decoded admin data and augmented vote data — they are enhancements. In the event that something goes wrong with them (in the serverless functions) we don't want to make the user think that the _whole_ app is broken by showing a "error fetching vote data" error.


daywiss (approved):
seems good to me

PR #123, opened 2022-12-02
fix: improve delegated commit/reveal flow

daywiss:

## Motiviation
we could not commit as a delgate. After fixing this bug it was apparent we could not reveal either.

## Summary
This slightly changes how we determine if we can commit/reveal but making 3 seperate queries to the vote contract to pull in committed votes. This needs to be done because there are multiple scenarios where depending on what wallet you are logged in as, you need to query the contract differently to find committed votes.  This covers all the scenearios (hopefully). 

This also adds a new field to the vote data called "canReveal", this tells you if the logged in account could potentially reveal this commit. This is important because a delegator can still vote even if they have delegated to another wallet.  This allows a delegator and delegate to know which votes they could reveal, and its possible to switch between wallets to do this now.

This has some UX impliciations which may need to be debated ( do we actually want to prevent delegator from committing/revealing), but this gives the most flexible solution for now.


ryanwolhuter (approved):
very solid work and well done for figuring this whole thing out so quickly

PR #124, opened 2022-12-09
fix: address issue in outstanding rewards

chrismaree:
This PR addresses a known issue with how unclaimed rewards are calculated in the front end to correctly track unclaimed rewards.


daywiss (approved):
lgtm

hooks/queries/rewards/useOutstandingRewards.ts
```diff
@@ -0,0 +1,23 @@
...
+import { useAccountDetails, useContractsContext, useHandleError } from "hooks";
+import { getOutstandingRewards } from "web3";
+
+export function useOutstandingRewards() {
+  const { voting } = useContractsContext();
+  const { address } = useAccountDetails();
+  const { onError } = useHandleError({ isDataFetching: true });
+
+  const queryResult = useQuery(
+    [outstandingRewardsKey, address],
```
ryanwolhuter, line 13: build is erroring because this `outstandingRewardsKey` doesn't exist. Did you add this to the `queryKeys.ts` file in the `constant` dir?

ryanwolhuter:
Looks like you nailed it. Only two comments:

Missing variable definition (see comment)

Seems like you didn't need to change the mutation code. I assume this all still works the same in terms of resetting the outstanding rewards and unstaked balance. Looks like it works all good, just want to make sure you've tested that.

ryanwolhuter (approved):

chrismaree:
> Looks like you nailed it. Only two comments:
> 
> Missing variable definition (see comment)
> 
> Seems like you didn't need to change the mutation code. I assume this all still works the same in terms of resetting the outstanding rewards and unstaked balance. Looks like it works all good, just want to make sure you've tested that.

Right! was not sure about how to fix this type definition on `queryKeys.ts`. thanks! regarding testing mutation I did test is a few times and it seemed to work more correctly now.

PR #125, opened 2022-12-09
fix: add process env for DISCORD_TOKEN

chrismaree:
enforce discord API token is added correctly in serverless function.


pages/api/discord-thread.ts
```diff
@@ -16,6 +16,8 @@ export async function discordRequest(
 ) {
   const url = "https://discord.com/api/v10/" + endpoint;
   if (options.body) options.body = JSON.stringify(options.body);
+  if (!process.env.DISCORD_TOKEN)
+    throw Error("DISCORD_TOKEN env variable not set!");
```
daywiss, line 20: its interesting this is thrown during function call rather than app load, but perhaps that causes issues with the serverless env

PR #126, opened 2022-12-14
fix: ensure we clear contracts with signers when it changes

daywiss:

This will clear any contracts which have a previous signer in it. this may help prevent the problem where we can possibly submit a transaction based on another wallet account.


ryanwolhuter (approved):
well done for catching this

PR #127, opened 2022-12-14
improve: add explicit signing button

daywiss:

## motiviation
When first opening the page metamask immediately asks us to sign a message. If this message is cancelled by the user, theres no clear way to re-sign without reloading the page. Would notice that multiple signature requests would be sent to users metamask. Additionally commit/reveal were not disabled if the user had not signed.

## fix
This gives you an explicit button to "sign"  with a tooltip before you can commit or reveal. We also refactor our button display logic to help make this more managable. 

image
image
image

## testing
to test this new flow, you must clear your application cache.
image

### DIFF
--- a/components/Votes/Votes.tsx
+++ b/components/Votes/Votes.tsx
@@ -7,10 +7,10 @@ import {
   VotesTableHeadings,
   VoteTimeline,
 } from "components";
+import { useConnectWallet } from "@web3-onboard/react";
 import { defaultResultsPerPage } from "constant";
 import { formatVotesToCommit, getEntriesForPage } from "helpers";
 import {
-  useAccountDetails,
   useCommitVotes,
   useContractsContext,
   useDelegationContext,
@@ -21,6 +21,7 @@ import {
   useVotesContext,
   useVoteTimingContext,
   useWalletContext,
+  useUserContext,
 } from "hooks";
 import { useState } from "react";
 import styled from "styled-components";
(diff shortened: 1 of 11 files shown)


components/Votes/Votes.tsx
```diff
@@ -50,26 +52,136 @@ export function Votes() {
...
-    (isCommit && !canCommit()) || (isReveal && !canReveal());
-  const disabledButtonExplanation = getDisabledButtonExplanation();
-  const buttonLabel = isCommit ? "Commit Votes" : "Reveal Votes";
-  const buttonOnClick = isCommit ? commitVotes : revealVotes;
   const votesToShow = getEntriesForPage(
     pageNumber,
     resultsPerPage,
     determineVotesToShow()
   );
+  const actionStatus = calculateActionStatus();
```
ryanwolhuter, line 60: much better thank you

contexts/UserContext.tsx
```diff
@@ -48,6 +56,8 @@ export const UserContext = createContext<UserContextState>(
 export function UserProvider({ children }: { children: ReactNode }) {
   const { connectedWallet, account, address, truncatedAddress } =
     useAccountDetails();
+  const { signingKeys } = useWalletContext();
```
ryanwolhuter, line 59: I should have done it like this from the start

ryanwolhuter (approved):
excellent work as always

daywiss:
merged this into the wrong branch

PR #128, opened 2022-12-15
feat: add discord panel styles

ryanwolhuter:
### Summary

Add a `Tab` in the `VotePanel` that shows the discord discussion for a given vote. This data comes from the `/api/discord-thread` serverless function.

This implementation is for styles only — dummy data is used. I have made the dummy data conform to the structure of the real data returned from the functions, see the type definitions in `types/voting.ts` for more details.


constant/voting/discord.ts
```diff
@@ -0,0 +1,3 @@
+export const discordLink = "https://discord.com/invite/jsb9XQJ";
+export const evidenceRationalDiscordChannelId = "964000735073284127";
```
daywiss, line 2: this shoudl probably also be an env, but not a blocker

daywiss (approved):

daywiss:
gonna go ahead and merge this, nice work

PR #130, opened 2022-12-16
improve: remove references to goerli, allow path to add chains

daywiss:
## motivation
we need to remove references to goerli from the app in order to prepare it to run on mainnet. 

## changes
this generalizes the app configuration, combines constants we need for each chain and selects it based on chain id specified in env.  i've tried to identify places in the app that were hard coded to goerli and replaced them with the generalized configuration. 

## important
This requires changes to vercels envs before this will run by removing references to goerli.  It also requires a new var that specifies the chain id the app is on.

### DIFF
--- a/helpers/config.ts
+++ b/helpers/config.ts
@@ -4,12 +4,11 @@ import * as ss from "superstruct";
 // and give them information as to which env was missing. We don't want the app to run without required variables.
 // This would also be a great place to document what each env does and how to find it.
 const Env = ss.object({
-  NEXT_PUBLIC_VOTING_V1_CONTRACT_ADDRESS_GOERLI: ss.string(),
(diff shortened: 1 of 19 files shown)


helpers/config.ts
```diff
@@ -78,17 +77,72 @@ export const appConfig = ss.create(
...
+
+export const chainConstantsList: ChainConstantsList = [
+  {
+    chainId: 1,
+    infuraName: "homestead",
+    properName: "Mainnet",
+    makeTransactionHashLink: (transactionHash: string) =>
+      `https://etherscan.io/tx/${transactionHash}`,
+    onboardConfig: {
+      id: "0x1",
```
chrismaree, line 117: strange that these are hex encoded.

chrismaree (approved):

PR #132, opened 2022-12-16
feat: add network switch button

daywiss:
## Motivation
Commit/reveal button was not being disabled when user was on wrong chain. 

## change
This changes the commit/reveal button behavior to allow the user to switch to the correct chain. 
image
This also removes the automatic behavior of immediately requesting metamask to switch chain, instead preferring the user explicitly click the button.

This is a slight issue becuase this button is hidden unlesss there is an active vote. will talk to jesper to see how we can resolve.


md0x (approved):
✨

PR #133, opened 2022-12-19
fix: number to total votes

md0x:
Signed-off-by: Pablo Maldonado <pablo@umaproject.org>

Fixes the total number of votes where the user participated in


daywiss (approved):
thanks for this!

PR #137, opened 2022-12-19
fix: make sure wallet connected is not red

daywiss:
when changing the connect button color to red, did not check the color once connected

image
fixed back to 
image


md0x (approved):
LGTM!

PR #139, opened 2022-12-21
feat: wire up discord api to discusstion panel

daywiss:

## Motivation
We had independently built a discussion panel for votes, and a discord serverless api.  We needed to wire these together to get discussions working correctly.

## Changes
This changes aspects of the serverless api. Previously the api requested an array of discord threads, but this design was not ideal for the way the frontend was designed to query. This was changed to query a single thread at a time, based on when the user opens the discussion panel ( rather than compiling a list of all possible discussions).  We also add superstruct to validate calls and results between the frontend and the serverless api. 

Several parsers were added to make mentions, links and attachments work somewhat more nicely. 

There is still a lot of refinements we can make on this.
1. mentions to other channels do not work correctly
2. images are not displayed, but rather are linked as attachments 
3. serverless function is not particularly efficient, it runs a big query to discord every time its hit
4. there is no loading icon on the discussion panel
5. discussion panel should show there has not been any discussion if no discussion exist
6. there should be a red dot or badge number showing how many messages are in the discussion tab ( before entering the tab)

image

### DIFF
--- a/components/Panel/VotePanel/Discussion.tsx
+++ b/components/Panel/VotePanel/Discussion.tsx
@@ -6,46 +6,15 @@ import NextImage from "next/image";
 import Message from "public/assets/icons/message.svg";
 import ReactMarkdown from "react-markdown";
 import styled, { css } from "styled-components";
-import { DiscordMessageT, DiscordThreadT, VoteT } from "types";
+import { VoteT } from "types";
 import { PanelSectionTitle } from "../styles";
+import { useDiscordThread } from "hooks";
 
 export function Discussion({ identifier, time }: VoteT) {
-  const dummyMessages: DiscordMessageT[] = [
-    {
-      message:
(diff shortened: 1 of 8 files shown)


chrismaree (approved):
I really like this! it works super well and the formatting is great! One bit of feedback from testing this is that we should render _nothing_ if there are no discussion threads for this particular request. either this means showing something here:

2022-12-28 at 08 55 17@2x

to say _there is no discussion for this request_ or we simply don't show the discussion tab.

other than this, this LGTM!

daywiss:
> I really like this! it works super well and the formatting is great! One bit of feedback from testing this is that we should render nothing if there are no discussion threads for this particular request. either this means showing something here:

@chrismaree I agree, id like to tackle it in a new issue which i've made here:  https://linear.app/uma/issue/UMA-955/voterdapp-improvement

PR #140, opened 2022-12-21
feat: enable real time apr display

daywiss:
## Motivation
previously we hardcoded the apy heading to 30% or something like that. we want to get the dynamic value.

## Changes
This uses the graph to pull the new value from "global" state.  This was recently updated so that we dont need user information to pull the apy value.  APY is also renamed to APR to match the how to section.


chrismaree (approved):
very nice!

PR #142, opened 2022-12-23
feat: add commit reset and dirty flag on input

daywiss:
## motivation
Previously when values were committed, it wasnt clear if we were changing them to be committed again.

## changes
This adds a button to reset inputs back to a previously committed vote, and lets you know if you are changing a previously committed value. Otherwise it does not change functionality
image


chrismaree (approved):
I tested this out on the staging link and it seems to work great!

PR #145, opened 2022-12-28
fix: fix delegate queries to use delegator address

daywiss:
# motivation
The current version does not work correctly when logged in as a delegate. Much of the data should originate from the delegator:
1. stake/unstake balances 
2. vote history
3. claimable tokens

# changes
This allows many contexts to set an override user address, in this case we can override to the "delegator" address. This happens automatically at the top level ( in header component) when its detected that you are a delegate. Many hooks have now an optional field to override the address, and contexts selectively do this when you set an override address. This fixes the issue without major changes to the code, and keeps it backwards compatible to lower risk of regressions. 

image
image

### DIFF
--- a/components/Header/Header.tsx
+++ b/components/Header/Header.tsx
@@ -1,3 +1,4 @@
+import { useEffect } from "react";
 import { Wallet } from "components";
 import {
   green,
@@ -11,6 +12,8 @@ import {
   useDelegationContext,
   usePanelContext,
   useStakingContext,
+  useUserContext,
+  useVotesContext,
 } from "hooks";
 import NextLink from "next/link";
 import Bell from "public/assets/icons/bell.svg";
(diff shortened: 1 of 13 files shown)


chrismaree:
I tested this out and it seems to work great! one bit of feedback is that we should change some of the copy to make it clear that you are connected via a delegate. in particular here:
2022-12-29 at 17 07 44@2x

we have the "connected as delegate wallet" in the top right which is good but this text should say "you are staking x of you y **delegate** UMA tokens".

chrismaree (approved):
this looks good! I just have one bit of UI feedback on the copy that we show the user. other than that I'm happy!

daywiss:
> this looks good! I just have one bit of UI feedback on the copy that we show the user. other than that I'm happy!

making an issue for this https://linear.app/uma/issue/UMA-967/improve-copy-around-how-to-amount-staked-as-delegate

PR #146, opened 2022-12-29
fix: filter past commit value from showing in a rolled vote

daywiss:
## motivation
When committing a vote, letting it roll to next round, the previous commit value keeps showing up as the default in the commit input box. The previous commit is not related to the next round when a vote is rolled, so it should be blank.

## changes
This was caused by not filtering decrypted votes by round id when doing a lookup for the committed value, this was treating past votes the same as votes in the current round. we slightly change queries to allow for a round id. this affected past votes, since they  were sharing the same data structure, so a new query is made for all past votes.


chrismaree (approved):
tested and this looks good!

PR #147, opened 2022-12-29
fix: ensure claim balance goes to 0 after claiming

daywiss:
## motivation
The amount to claim does not go to 0 after claiming rewards, seems to require a reload or takes a long time to reflect new value after tx is mined.

## changes
this was caused by a race condition, where claims are being regenerated on a timer based on contract information. resetting this calculated value to 0 would be quickly reset by the timer since the contract value has not been updated yet.  rather than resetting the final value to 0 when the claim tx succeeds, we change the cached value of the outstanding rewards from the contract call to 0, which is the key value in calculating our claim amount and wont immediately be overridden by the timer.

### DIFF
--- a/contexts/StakingContext.tsx
+++ b/contexts/StakingContext.tsx
@@ -26,7 +26,6 @@ export interface StakingContextState {
   canUnstakeTime: Date | undefined;
   unstakeCoolDown: BigNumber | undefined;
   v1Rewards: V1RewardsT | undefined;
-  resetOutstandingRewards: () => void;
   getStakingDataLoading: () => boolean;
   getStakingDataFetching: () => boolean;
   setAddressOverride: (address?: string) => void;
@@ -42,7 +41,6 @@ export const defaultStakingContextState: StakingContextState = {
   unstakeRequestTime: undefined,
   canUnstakeTime: undefined,
   unstakeCoolDown: undefined,
-  resetOutstandingRewards: () => null,
   v1Rewards: undefined,
   getStakingDataLoading: () => false,
   getStakingDataFetching: () => false,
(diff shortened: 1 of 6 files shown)


chrismaree:
I tested this on the staging link it it did not seem to correctly fix this issue. I clicked the "claim and re-stake button" and did nothing until the tx mined. once it mined the UI did not update to show me that my claimable rewards was zero, nor did my balance go up.

going to test a second time now

chrismaree:
> I tested this on the staging link it it did not seem to correctly fix this issue. I clicked the "claim and re-stake button" and did nothing until the tx mined. once it mined the UI did not update to show me that my claimable rewards was zero, nor did my balance go up.
> 
> going to test a second time now

it looks like it worked when claiming to my wallet but not when I claim & re-stake. ideally this works with boths kinds of claims.

daywiss:
> it looks like it worked when claiming to my wallet but not when I claim & re-stake. ideally this works with boths kinds of claims.

@chrismaree Ok so it looks like i needed to reset a timer as well to properly reset the value, this is in for both withdraw and withdraw and stake. I've tried it a few times and it seems ok, but theres some wierdness around react/next that causes the estimated reward value to bounce around which i may have to correct in another pr.

chrismaree:
> > it looks like it worked when claiming to my wallet but not when I claim & re-stake. ideally this works with boths kinds of claims.
> 
> @chrismaree Ok so it looks like i needed to reset a timer as well to properly reset the value, this is in for both withdraw and withdraw and stake. I've tried it a few times and it seems ok, but theres some wierdness around react/next that causes the estimated reward value to bounce around which i may have to correct in another pr.

testing again now.

chrismaree (approved):
This looks like it's working fine now

PR #149, opened 2022-12-29
fix: allow app to work without graph or contentful

daywiss:
## Motivation
We want to be able to run the app in the absense of contentful or the graph. The app should fail gracefully if either app is configured to not include these, or these endpoints fail to return data. 

## Changes
The app can now be explicitly configured to not include these services by making their environment variable configurations optional.  By disabling these we no longer have access to certain features such as apy calculations, or past vote history, but the app stays stable.  If these endpoints error or timeout the app will report the errors to the user but still work.


ryanwolhuter (approved):
great work as always

PR #151, opened 2023-01-03
fix: fix magic number crash and display

daywiss:
## motivation
There were a couple issues, when viewing the zelensky vote, the app would crash due to magic number breaking display formatting. once that was fixed, it would show 2 no values, because the magic number value on the vote was not matching the value on the option, this ended up showing 2 no values rather than yes, no, early vote.

## changes
the root cause of this problem was pulling the correct vote number from the subgraph and immediately converting it to a number, this ends up converting the magic number to something that cannot be converted back to the original number.  to fix this we stop tryign to convert votes into numbers and leave them as raw strings. when the number is finally displayed its converted to the correct decimals. 

## QA
go to page 2 of past votes, and see the zelensky time magazine vote. open the details panel and see the yes/no/early vote chart working correctly, app not crashing

### DIFF
--- a/helpers/voting/makePriceRequestsByKey.ts
+++ b/helpers/voting/makePriceRequestsByKey.ts
@@ -1,5 +1,11 @@
 import { decodeHexString, makeUniqueKeyForVote } from "helpers";
-import { PriceRequestByKeyT, PriceRequestT, RawPriceRequestDataT } from "types";
+import {
+  PriceRequestByKeyT,
+  PriceRequestT,
+  RawPriceRequestDataT,
+  VoteParticipationT,
+  VoteResultsT,
+} from "types";
 
 export function makePriceRequestsByKey(
   priceRequests: RawPriceRequestDataT[] | undefined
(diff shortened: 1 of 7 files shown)


helpers/voting/makePriceRequestsByKey.ts
```diff
@@ -22,7 +28,9 @@ function formatPriceRequests(priceRequests: RawPriceRequestDataT[]) {
     .filter((priceRequest): priceRequest is PriceRequestT => !!priceRequest);
 }
 
-function formatPriceRequest(priceRequest: RawPriceRequestDataT) {
+function formatPriceRequest(
+  priceRequest: RawPriceRequestDataT
+): PriceRequestT & VoteParticipationT & VoteResultsT {
```
daywiss, line 33: not related, but we return the correct type here rather than casting previously

types/voting.ts
```diff
@@ -22,7 +22,7 @@ export type PriceRequestT = {
   identifier: string;
   ancillaryData: string;
   voteNumber: BigNumber | undefined;
-  correctVote?: number;
+  correctVote?: string;
```
daywiss, line 25: votes are now raw strings rather than numbers so we dont lose information when the number is really large

ryanwolhuter (approved):
I wish I did it like this right from the start, would have saved a lot of headaches

PR #152, opened 2023-01-05
feat: add staking and migration guide banner

ryanwolhuter:
### Summary

Implements the banner as per the designs. When the user hides the banner, this is persisted in local storage.

I'll [update the link](https://linear.app/uma/issue/UMA-975/add-link-to-instructions-once-they-are-published) when the medium article is published.

Screenshot 2023-01-05 at 09 05 12

Screenshot 2023-01-05 at 09 05 24

Screenshot 2023-01-05 at 09 05 35


daywiss (changes requested):
just tested this, banner went away the first time, but reloaded and it came back, then i could not get rid of it. maybe something with this local storage library or the way its being hidden?

ryanwolhuter:
> just tested this, banner went away the first time, but reloaded and it came back, then i could not get rid of it. maybe something with this local storage library or the way its being hidden?

woops! this was an ssr issue. should be fixed now with the latest change

daywiss (approved):

PR #153, opened 2023-01-05
improve: allow augment request to configure fromblocks

daywiss:
## Motivitation
Consistently getting errors in the serverless function due to large event queries, this is preventing data from being returned from the augment request calls. 

## changes
The initial solution was to add a way to configure block start times for each event query, but the way the types and code was structured it was quite difficult to get typescript to approve.  used this as an opportunity to improve the code:
1. optional block start times per contract per chain are added, currently defaulting to 0
2. less casting, stronger type checks generally
3. data parameters sent to serverless function is now validated with superstruct before use
4. an optional parameter for chain id is now available to the caller, rather than hard coding this to 1 ( defaults to 1 if omitted)
5. data returned from serverless function is more strongly validated on the frontend with super struct, though it still gets cast in the end
6. structure of serverless code is updated to query data at the beginning of the call, rather than throughout the code to make it easier to understand and optimize calls if needed
7. event data returned from queries are treated as unknown, and put into a common event type, which gets validated with superstruct
8. efficiency is improved from n*m lookups to n+m where n is the number of requests to augment, and m is the number of events returned from queries
9. Promise.allSettled is used instead of Promise.all when making multiple event queries, to prevent a single failure from stopping calculations
10. optional debugging output with new env `DEBUG`
11. more obvious way to enable/disable querying of certain contracts at the top of the file

These changes had the effect of eliminating the initial problem without needing to change the startblock times ( all still 0). its possible perhaps an issue was causing the request to fail and keep requering, but not exactly sure why this fixed it. 

sample of logged data in frontend
image

### DIFF
--- a/pages/api/_common.ts
+++ b/pages/api/_common.ts
@@ -36,8 +36,8 @@ export function isSupportedChainId(
   return chainId in supportedChains;
 }
 
-// add block start times here, can also convert this to envs somehow. this is used for augmented request evnets
-// and can be used in other places as well to optimize event calls.
+// add block start times here, perhaps we can allow this to be configured through environment variables eventually.
+// this is used for augmented request events and can be used in other places as well to optimize event calls.
 export const fromBlocks: Record<string, Record<number | string, number>> = {
   OptimisticOracle: {
     1: 0,
--- a/pages/api/_common.ts
+++ b/pages/api/_common.ts
@@ -1,12 +1,13 @@
 import { getAbi, getAddress } from "@uma/contracts-node";
 import { Contract, ethers } from "ethers";
 import { NodeUrls, SupportedChainIdsWithGoerli } from "types";
+import { supportedChains } from "constant";
 
 type GetAddressParams = Parameters<typeof getAddress>;
 
 type GetAbiParams = Parameters<typeof getAbi>;
 
-type ContractName = GetAddressParams[0] & GetAbiParams[0];
+export type ContractName = GetAddressParams[0] & GetAbiParams[0];
 
 export function getProviderByChainId(chainId: SupportedChainIdsWithGoerli) {
   return new ethers.providers.JsonRpcBatchProvider(getNodeUrls()[chainId]);
@@ -26,4 +27,44 @@ export async function constructContract(
     getAbi(contractName),
     getProviderByChainId(chainId)
   );
+}
+
+export function isSupportedChainId(
+  chainId: string | number | undefined
+): chainId is SupportedChainIdsWithGoerli {
+  if (chainId === undefined) return false;
+  return chainId in supportedChains;
+}
+
+// add block start times here, can also convert this to envs somehow. this is used for augmented request evnets
+// and can be used in other places as well to optimize event calls.
+export const fromBlocks: Record<string, Record<number | string, number>> = {
+  OptimisticOracle: {
+    1: 0,
+    5: 0,
+  },
+  OptimisticOracleV2: {
+    1: 0,
+    5: 0,
+  },
+  SkinnyOptimisticOracle: {
+    1: 0,
+    5: 0,
+  },
+  VotingV2: {
+    1: 0,
+    5: 0,
+  },
+  Voting: {
+    1: 0,
+    5: 0,
+  },
+};
+
+export function getFromBlock(
+  oracleType: string,
+  chainId: number,
+  fallbackFromBlock = 0
+): number {
+  return fromBlocks?.[oracleType]?.[chainId] || fallbackFromBlock;
 }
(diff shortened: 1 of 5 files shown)


pages/api/augment-request.ts
```diff
@@ -38,148 +77,156 @@ function castOracleNameForOOUi(oracleType: string): string {
...
-        ? optimisticOracle.queryFilter(optimisticOracle.filters.RequestPrice())
-        : null
+  return events.map((event) =>
+    ss.create(
+      {
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
+        transactionHash: event?.transactionHash || "rolled",
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
+        identifier: event?.args?.identifier,
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
```
daywiss, line 115: the linter really hated this, even though its pretty safe. we just need to drill into an unknown object, and then use superstruct to validate the end result, this means we dont need any explicit casting, so i disabled all the linter errors

pages/api/_common.ts
```diff
@@ -27,3 +28,43 @@ export async function constructContract(
...
 }
+
+export function isSupportedChainId(
+  chainId: string | number | undefined
+): chainId is SupportedChainIdsWithGoerli {
+  if (chainId === undefined) return false;
+  return chainId in supportedChains;
+}
+
+// add block start times here, can also convert this to envs somehow. this is used for augmented request evnets
```
ryanwolhuter, line 39: typo "evnets"

pages/api/augment-request.ts
```diff
@@ -161,7 +165,10 @@ async function getOracleRequestPrices(
         // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
         identifier: event?.args?.identifier,
         // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
-        time: event?.args?.timestamp,
+        time:
+          event?.args?.timestamp instanceof BigNumber
+            ? event?.args?.timestamp?.toNumber()
+            : event?.args?.timestamp,
```
daywiss, line 171: somehow the events are different, sometimes its a number sometimes its a big number, this will make that check

ryanwolhuter (approved):
this is really excellent work dude well done! I'm approving but I'd wait for chris before merging

pages/api/augment-request.ts
```diff
@@ -38,148 +78,162 @@ function castOracleNameForOOUi(oracleType: string): string {
...
+  contractType: VotingType,
   chainId: SupportedChainIds
-) {
-  const chainOptimisticOracles = await constructOptimisticOraclesByChain(
-    chainId
+): Promise<CommonEventData[]> {
+  const contract = await constructContract(chainId, contractType);
+  const events = await contract.queryFilter(
+    contract.filters.PriceRequestAdded(),
+    getFromBlock(contractType as string, chainId as number)
```
chrismaree, line 107: does changing the `fromBlock` actually modify this if the difference is 0 or the deployment block? so say that a contract is deployed at block 100. does it actually save request size and time to set the value in a `queryFilter` to set the value to 100, as apposed to 0? I thought that it just shortcut given how the event lookup works in infura.

types/voting.ts
```diff
@@ -220,6 +220,24 @@ export type OracleTypeT =
   | "OptimisticOracleV2"
   | "SkinnyOptimisticOracle";
 
+export const AugmentedVoteDataResponseT = ss.object({
```
chrismaree, line 223: this is useful

chrismaree (approved):
this looks great! good work. only feedback is that I think some of this modification of fromblock stuff might not actually directly save query size, if its only from the deployment block.

at a later point we can look into building a more robust caching system as we only really need to re-process augment requests when a new request is produced, and not re-do all processing every time a single one happens.

pages/api/augment-request.ts
daywiss: i didnt know how intelligent event queries were, i assumed the worst, it would be great if he started from contract deployment automatically. would make this feature redundant, but technically you could bump this configuration after the deploy time if you didn't care about data over x days old.  i havent seen any deterioration in behavior since this change though, so I dont think its any worse than before.

pages/api/augment-request.ts
daywiss: its a little more specific this way, rather than turning it off for the entire file, its only for these lines, but good to know that is an option 😄

PR #154, opened 2023-01-05
improve: remove references to goerli as a supported chain, include it natively

daywiss:
## motivation
We dont want to special case the testnet if we dont have to, doing this has also caused typescript issues when trying to find names for supported chains when using goerli, as that object did not have a goerli id.  

## changes
this simply adds goerli into supported chain object and supported chain type, and removes references to the type with goerli.

**do not merge until #153 goes in, the base needs to be set to master**

### DIFF
--- a/constant/web3/supportedChains.ts
+++ b/constant/web3/supportedChains.ts
@@ -1,5 +1,6 @@
 export const supportedChains = {
   1: "Ethereum",
+  5: "Görli",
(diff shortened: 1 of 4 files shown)


ryanwolhuter (approved):
yeah this is definitely cleaner

chrismaree (approved):

PR #155, opened 2023-01-05
fix: show correct etherscan link for chain

daywiss:
## Motivation
Since fixing the augmented data, its clear that some of the links are not linking to the correct etherscan site, its set to only link to the primary chain ids etherscan.  testing this more, we also see that polygon requests are also not showing correctly.

## changes
This generalizes the transaction link creation logic to accept any chain, and when creating the details links, supplies the chain id. chainid is still optional, so if not supplied it will default to the apps primary chain.  This also adds polygon constant data so that we can generate correct links to transactions. 

image

**do not merge, #983 must go in first and change base to master**


ryanwolhuter (approved):
nice one

chrismaree (approved):

PR #156, opened 2023-01-05
fix: get goerli augment data to show up

daywiss:
## motivation
when testing on goerli, there are a few issues with the augmented request data
1. links to testnet oo ui are not wokring correctly since the are linking to mainnet
2. augmented data default chain id is hardcoded to 1 rather than the configured primary chain

## changes
We add the ability to switch urls between mainnet and testnet so augmented oo links work. We also default to our primary chain id, which will change depending on mainnet/testnet deployment. Some other minor changes to reduce casting, fix some conventions and bump contract dependencies.


chrismaree (approved):
cool!

PR #157, opened 2023-01-06
feat: add why must I sign tooltip

ryanwolhuter:
Signed-off-by: ryanwolhuter <dev@ryanwolhuter.com>

Add a simple tooltip that explains why we need a user's signature. Likely will need a copy update, but the placeholder text is good enough for now. The tooltip only shows when the commit / reveal action button is "Sign".

### DIFF
--- a/components/Votes/Votes.tsx
+++ b/components/Votes/Votes.tsx
@@ -1,5 +1,7 @@
+import { useConnectWallet } from "@web3-onboard/react";
 import {
   Button,
+  IconWrapper,
   Pagination,
   Tooltip,
   VotesList,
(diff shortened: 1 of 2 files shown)


components/Votes/Votes.tsx
```diff
@@ -212,6 +214,10 @@ export function Votes() {
     }
     return actionConfig;
   }
+
+  const signatureExplanationText =
+    "UMA uses this signature to verify that you are the owner of this address. We must do this to prevent double voting.";
```
daywiss, line 219: "UMA uses this signature to verify that you are the owner of this address. This allows us to hide your vote during the commit phase and decode it during the reveal phase."

@mrice32 @chrismaree might want to double check this copy

daywiss (changes requested):

components/Votes/Votes.tsx
chrismaree: ye, we dont need to worry too much about the copy as @smb2796 has this other copy https://github.com/UMAprotocol/voter-dapp-v2/pull/138 PR that we aim to fix all the copy in one go. that'll be worked on this week.

chrismaree (approved):

daywiss (approved):

PR #158, opened 2023-01-06
fix: hide unstake as delegate

daywiss:
## motivation
if your delegator unstakes, and has pending unstake, only that account can claim it. the delegate should not see that there is pending unstaked tokens.

## changes
this hides the pending unstake if you are a delegate
delegate view ( account 8):
image

delegator view (account 3):
image


chrismaree:
would it be useful to add some text to this part of the panel to say why you cant unstake? or explain the relationship between delegator/delegate to say that this is something you cant do and need to use the other wallet? even a link to the docs would be good, I think.

ryanwolhuter (approved):
perfect

daywiss:
> would it be useful to add some text to this part of the panel to say why you cant unstake? or explain the relationship between delegator/delegate to say that this is something you cant do and need to use the other wallet? even a link to the docs would be good, I think.

@chrismaree personally i think just hiding it is better. as a delegate i dont think you even need to know what the delegator is doing on their account, ie staking/unstaking.  but maybe we should also get opinion from @smb2796

PR #159, opened 2023-01-06
fix: fixed blocked commits, and delegate/delegator statuses

daywiss:
## motivation
turns out when you have a delegate/delegator relationship, and both commit 1 vote, you can no longer commit again for any votes you have not already committed. 

## changes
This fixes the logic disabling the commit button incorrectly, the logic is a bit tricky, but it seems to work now. also used this as an opportunity to clarify labels when votes are committed from delegate/delegator account. probably should have put this in a new PR but we can see the changes here:
commit cycle where delegator committed viewing from delegate:

reveal cycle from various viewpoints, and a reveal

### DIFF
--- a/components/Votes/Votes.tsx
+++ b/components/Votes/Votes.tsx
@@ -96,8 +96,9 @@ export function Votes() {
     const isDelegate = getDelegationStatus() === "delegate";
     const hasSigner = !!signer;
     const votesToShow = determineVotesToShow();
-    const hasPreviouslyCommitted =
-      votesToShow.filter((vote) => vote.decryptedVote).length > 0;
+    const hasPreviouslyCommittedAll =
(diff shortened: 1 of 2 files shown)


daywiss:
> Are the third and fourth screenshots also from the delegate's point of view? I ask because they say delegate must reveal and I wonder if instead there should only be an explicit definition of delegator/delegate must do something if it is not the role that they have.

the last 2 screens are from the delegators point of view, the votes were committed by the delegate.  the logic will check if you are delegate/delegator and set the correct terminology accordingly, i.e. if you are a delegate, it will say the delegator needs to reveal

smb2796:
Cool, makes sense then!

smb2796 (approved):

ryanwolhuter (approved):
yeah this is tricky, well done for solving it simply

PR #162, opened 2023-01-09
fix: corrected dvm links in details

daywiss:
## motivation
we are labeling links as "DVM requests" but then linking to the OO transaction, this seems incorrect. There should probably be 2 links, 1 for the dvm and 1 for the OO ( if its available).

## changes
This fixes the link to dvm so it correctly is based on the dvm price request event transaction. Then adds an optional second link which is based on the OO request event, which may or may not exist for a request.

before:
image
image

after:
image
image

### DIFF
--- a/constant/web3/supportedChains.ts
+++ b/constant/web3/supportedChains.ts
@@ -10,15 +10,16 @@ export const supportedChains = {
   42161: "Arbitrum",
 };
 
-export function getOracleProperName(oracleType:string):string{
+export function getOracleTypeDisplayName(oracleType:string | undefined) {
+  if(!oracleType) return "Optimistic Oracle v1";
   switch (oracleType) {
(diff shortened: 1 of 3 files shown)


constant/web3/supportedChains.ts
```diff
@@ -9,3 +9,16 @@ export const supportedChains = {
   43114: "Avalanche",
   42161: "Arbitrum",
 };
+
+export function getOracleProperName(oracleType:string):string{
```
ryanwolhuter, line 13: nit: spacing and redundant return type

ryanwolhuter (approved):
👍

constant/web3/supportedChains.ts
daywiss: thats wierd, why is the linter not picking that up

PR #163, opened 2023-01-09
feat: rank active votes by spam deletion and vote number

daywiss:
## motivation
we want spam deletion proposal votes during an active vote to be prioritized at the top of the list.

## changes
this will order spam deletion proposals to the top of the active votes list. it also sortes by vote number and timestamp ascending which changes the old behavior

## qa
this cant be tested currently until we have some spam deletion proposals. will wait until we have some to merge


daywiss:
span deletion no longer exist

PR #164, opened 2023-01-09
improve: change wording for delegate staking amount

daywiss:
## motivation
we want a delegate to know that its not thier tokens being staked, but the delegators in the "how it works" section.

## changes
this changes the wording slightly to show its the delegators uma at stake, and links to the wallet connect page to see your delegator. 

image
image

@ryanwolhuter there is supposed to be a tooltip here, but i cant seeem to get it working, let me know how i can fix

### DIFF
--- a/components/HowItWorks/HowItWorks.tsx
+++ b/components/HowItWorks/HowItWorks.tsx
@@ -80,10 +80,10 @@ export function HowItWorks() {
                       : ""
                   }
                 >
-                  <>
+                  <span>
                     <Link href="/wallet-settings">Your delegator</Link> is
                     staking
-                  </>
+                  </span>
                 </Tooltip>
               ) : (
                 "You are staking"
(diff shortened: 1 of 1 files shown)


components/HowItWorks/HowItWorks.tsx
```diff
@@ -63,15 +72,30 @@ export function HowItWorks() {
...
-              You are staking{" "}
+              {isDelegate ? (
+                <Tooltip
+                  label={
+                    delegatorAddress
+                      ? `Delegator address ${delegatorAddress}`
+                      : ""
+                  }
+                >
+                  <>
```
ryanwolhuter, line 83: wrap this `<Link>` in a `<span>` instead of a fragment to fix the issue. the tooltip wants a real a dom node, breaks with the virtual one from the fragment

ryanwolhuter (approved):
see comment for solution for the tooltip. tried it out on my local and it works

components/HowItWorks/HowItWorks.tsx
daywiss: awesome will try

daywiss: that fixed it!

PR #165, opened 2023-01-10
improve: integrate new voting v2 contract

daywiss:
## motivation
we want to integrate the updated voting v2 contract

## changes
this just changes the variables and events that have changed name


ryanwolhuter (approved):
lgtm

PR #168, opened 2023-01-13
feat: add recent past votes to vote page

ryanwolhuter:
Adds a "recent past votes" section to the bottom of the vote page, with a link to see them all at `/past-votes`.

Screenshot 2023-01-13 at 15 05 07

### DIFF
--- a/components/Header/Header.tsx
+++ b/components/Header/Header.tsx
@@ -292,7 +292,7 @@ const NotificationIconWrapper = styled.div`
   }
 `;
 
-const Dot = styled.div`
+const Dot = styled.span`
   height: 8px;
   width: 8px;
   background-color: ${green};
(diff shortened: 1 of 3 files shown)


components/Header/Header.tsx
```diff
@@ -292,7 +292,7 @@ const NotificationIconWrapper = styled.div`
   }
 `;
 
-const Dot = styled.div`
+const Dot = styled.span`
```
ryanwolhuter, line 295: these divs were resulting in invalid html because div cannot be a child of p. Changed to span to fix, functionality is identical

daywiss:
seeing a weird thing, dont know what pr this started in 
image
will make an issue for it unless u think it should be resolved here

PR #169, opened 2023-01-13
improve: update notification icons from design

ryanwolhuter:
Use the new notification icons from the design.


daywiss (approved):
coudl use some screens to see this in action

PR #170, opened 2023-01-13
fix: disable unstaking during request and countdown

daywiss:

## motivation
two things here, we could send multiple unstake requests, which woudl cause reversions in all but one transaction. second when cooling down, we disable buttons without a warning as to why, this should be identical to whne cooldown ends and it says " you cannot unstake until you claim unstaked tokens".

## changes
This passes some new data into the unstake panel which lets us check if theres a pending request to unstake, as wel as if we are in cooldown status, and disables/adds warnings accordingly. 

image
image

### DIFF
--- a/components/Panel/StakeUnstakePanel/StakeUnstakePanel.tsx
+++ b/components/Panel/StakeUnstakePanel/StakeUnstakePanel.tsx
@@ -112,6 +112,8 @@ export function StakeUnstakePanel() {
           unstakeCoolDown={unstakeCoolDown}
           isReadyToUnstake={isReadyToUnstake}
           isDelegate={isDelegate}
+          hasCooldownTimeRemaining={hasCooldownTimeRemaining}
+          isRequestingUnstake={isRequestingUnstake}
         />
       ),
     },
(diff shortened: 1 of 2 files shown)


ryanwolhuter (approved):
makes sense

smb2796:
Looks great! Is it possible to have the "in cool-down period" warning message be separate than what shows up in situation where "cooldown is over, need to claim unstaked tokens"? Looks like it is reusing the same warning rn, and I'd like to have the copy be slightly different btwn the two.

daywiss:
> Looks great! Is it possible to have the "in cool-down period" warning message be separate than what shows up in situation where "cooldown is over, need to claim unstaked tokens"? Looks like it is reusing the same warning rn, and I'd like to have the copy be slightly different btwn the two.

@smb2796  Can do, did you have specific copy in mind? in the mean time ill just add something that makes sensee

smb2796:
> > Looks great! Is it possible to have the "in cool-down period" warning message be separate than what shows up in situation where "cooldown is over, need to claim unstaked tokens"? Looks like it is reusing the same warning rn, and I'd like to have the copy be slightly different btwn the two.
> 
> @smb2796 Can do, did you have specific copy in mind? in the mean time ill just add something that makes sensee

I was updating the copy in #166 for this page. In state where it is not the cool-down timer, I'd like it to be:

"You cannot request to unstake until you claim your previously unstaked tokens."

During the cooldown timer, I'd like it to be:

"You cannot request to unstake additional tokens until you wait X time for your previous unstake request to complete."

PR #171, opened 2023-01-13
fix: incorrect social icon names

daywiss:
## motivation
some links and social icons were not matched correctly, discourse and medium.

## changes
this renames the icons and imports them to the correct links. there were a couple side effects of this, there is no longer an email link, as that icon was actually the discourse icon, and this changes the discord chat integration, as it was incorrectly using the discourse icon. 

before:
image

after: 
image


ryanwolhuter (approved):
Turns out I don't know my social media logos very well lol

smb2796 (approved):
Looks good!

PR #173, opened 2023-01-17
fix: incorporate new voter contract

daywiss:

## motivation
there are slight changes in teh new voter contract abi, we need to update app to use

## changes
this only changes the basics needed to get this to run correctly. there will be follow ups to make use of the new features


ryanwolhuter (approved):
nice work

daywiss:
there seems to be an issue with committing that we are trying to resolve. unclear if its a frontend or contract issue. will merge this and patch later if needed.

PR #174, opened 2023-01-17
improve: move roll logic out of severless

daywiss:
## motivation
new contract keeps track of roll count, so we dont need to do this bsaed on the augmented data api call

## changes
this uses the rollcount number returned from get active requests
image

## must change base from #173 after its merged


ryanwolhuter (approved):
👍

PR #175, opened 2023-01-18
feat: show message when user is recommitting a vote

ryanwolhuter:
### Motivation

The user should be informed that when they want to change an already-committed vote, they will need to perform a new commit transaction.

### Changes

* Updated the dirty inputs data structure to know whether a given input is a re-commit of an existing vote
* Show an asterisk next to a vote that is being recommitted
* Show a message that says "Changes to committed votes need to be re-committed" if any dirty inputs are re-committing


ryanwolhuter:
@daywiss let me know if this implementation is okay. I considered adding the re-committing info to the action status, but then we would have had to have another field separate from the `infoText` one — because this message is displayed differently. I figured just using the boolean to show the message makes sense in this context, but I'm keen to hear your feedback.

daywiss:
@ryanwolhuter i know you refactored quite a bit, but that logic was really touchy, and it caused some incorrect behavior, went ahead and just minimized the changes needed.

components/Votes/Votes.tsx
```diff
@@ -431,17 +436,21 @@ const Title = styled.h1`
 
 const ButtonWrapper = styled.div`
   display: flex;
-  flex-direction: column;
-  align-items: end;
-  justify-content: center;
+  justify-content: end;
```
daywiss, line 439: something happened to this button wrapper which was putting the buttons in a column rather than a row. tried to correct it here, might want to double chekc this work

components/Votes/Votes.tsx
```diff
@@ -335,6 +335,11 @@ export function Votes() {
           ))}
         />
       </VotesTableWrapper>
+      {isDirty() ? (
+        <RecommittingVotesMessage>
+          * Changes to committed votes need to be re-committed
+        </RecommittingVotesMessage>
```
daywiss, line 341: this was the only change needed afaik to do this, though i do realize the dirty logic isnt implemented the best, its just really fragile

ryanwolhuter: I must have misunderstood what the dirty logic does. What exactly does it mean for the input to be dirty? I thought that inputs are dirty if they have any input, not only if they are changing an existing value.

daywiss: you are probably right, but in this case its if the value has changed from a previous commit. Should have commented it

daywiss (approved):

PR #178, opened 2023-01-19
fix: show resolved vote numbers in details page

daywiss:

## motivation
vote numbers are only available on resolved requests. its now known as `resolvedPriceRequestIndex`

## changes
this uses `resolvedPriceRequestIndex` when its available
image
image
image


ryanwolhuter (approved):
👍

PR #179, opened 2023-02-01
fix: fix revealed issue for delegate

daywiss:
## motivation
the ui would not show delegates who commmitted and revealed

## changes
this fixes it so it correctly detects when a delegate has revealed by using the delegators address when querying for revealed events.  this also should show you when logged in as a delegate or delegator if your sibling address has revealed

### DIFF
--- a/components/VotesList/VotesListItem.tsx
+++ b/components/VotesList/VotesListItem.tsx
@@ -210,9 +210,17 @@ export function VotesListItem({
       if (!isCommitted) return "Not committed";
       if (!decryptedVote || !canReveal) {
         if (delegationStatus === "delegator") {
-          return "Delegate Must Reveal";
+          if (isRevealed) {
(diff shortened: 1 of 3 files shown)


components/VotesList/VotesListItem.tsx
```diff
@@ -210,9 +210,17 @@ export function VotesListItem({
       if (!isCommitted) return "Not committed";
       if (!decryptedVote || !canReveal) {
         if (delegationStatus === "delegator") {
-          return "Delegate Must Reveal";
+          if (isRevealed) {
+            return "Delegate Revealed";
```
ryanwolhuter, line 214: tiny nit: let's not capitalize each word to stay consistent with the rest of the text

ryanwolhuter (approved):
lgtm 💪

PR #180, opened 2023-02-02
fix: discussion tab

ryanwolhuter:
### Motivation

fix the issues raised in:

https://linear.app/uma/issue/UMA-1108/render-nothing-if-we-have-no-discord-discussion
https://linear.app/uma/issue/UMA-1110/fix-strange-wrapping-of-text-in-discussion-tab
https://linear.app/uma/issue/UMA-1115/while-in-upcoming-vote-phase-and-while-in-commit-vote-phase-the-dapp

### Changes

for 1108:
* add "no discussion" message for empty discussion thread state
* add loading spinner to discussion tab
for 1110:
* use `break-word` for paragraph tags in rendered markdown
for 1115:
* always show the `Details` and `Discussion` tabs in the `VotePanel`
other:
* removed redundant `PanelContentT` type that caused an an error when the vote passed in is empty
* renamed `DiscordThreadT` types to `VoteDiscussionT` to prevent confusing `thread.thread` situations and to clarify that the data type contains both the data that identifies the vote (`identifier` and `time`) as well as the `thread` from discord


components/Panel/VotePanel/Discussion.tsx
```diff
@@ -32,42 +33,60 @@ export function Discussion({ identifier, time }: VoteT) {
...
-                  }}
-                >
-                  {message}
-                </ReactMarkdown>
-              </MessageTextWrapper>
-            </MessageContentWrapper>
-          </MessageWrapper>
-        </SectionWrapper>
-      ))}
+      {loading ? (
```
ryanwolhuter, line 36: this is all the same, just wrapped in a loading check

components/Panel/VotePanel/Discussion.tsx
```diff
@@ -83,6 +102,7 @@ export function Discussion({ identifier, time }: VoteT) {
 const Wrapper = styled.div`
   margin-top: 20px;
   padding-inline: 30px;
+  width: calc(var(--panel-width) - 15px);
```
ryanwolhuter, line 105: absolute width to fix word wrap

components/Panel/VotePanel/Discussion.tsx
```diff
@@ -132,13 +152,17 @@ const handleWordBreak = css`
...
-  ${handleWordBreak}
   font: var(--text-md);
   &:not(:last-child) {
     margin-bottom: 15px;
   }
 `;
 
+const MessageText = styled(Text)`
+  word-wrap: break-word;
+  word-break: break-word;
```
ryanwolhuter, line 163: break word instead of break all on paragraph tags in the markdown, also to fix  word wrap

components/Panel/VotePanel/VotePanel.tsx
```diff
@@ -8,12 +9,12 @@ import { Discussion } from "./Discussion";
 import { Result } from "./Result";
 
 interface Props {
-  content: PanelContentT | undefined;
+  content: VoteT;
```
ryanwolhuter, line 12: I changed the name of this prop because it really just takes a vote and there's no need to have an alias type

components/Panel/VotePanel/VotePanel.tsx
```diff
@@ -24,29 +25,49 @@ export function VotePanel({ content }: Props) {
...
-      content: (
-        <Result
-          decodedIdentifier={decodedIdentifier}
-          participation={participation}
-          results={results}
-          options={options}
-        />
-      ),
-    },
+  const { data: discussion, isFetching: discussionLoading } = useVoteDiscussion(
```
ryanwolhuter, line 28: fetch this data here so that we can load it so long as soon as the user opens the panel

hooks/index.ts
```diff
@@ -16,6 +16,7 @@ export { useInterval } from "./helpers/useInterval";
 export { useIsomorphicLayoutEffect } from "./helpers/useIsomorphicLayoutEffect";
 export { useMounted } from "./helpers/useMounted";
 export { usePanelWidth } from "./helpers/usePanelWidth";
+export { useSign } from "./helpers/useSign";
```
ryanwolhuter, line 19: my editor organized these in alphabetical order, I guess its nice so left it like that

hooks/queries/votes/useVoteDiscussion.ts
```diff
@@ -0,0 +1,17 @@
+import { useQuery } from "@tanstack/react-query";
```
ryanwolhuter, line 1: Renamed to avoid confusion between discussion and the thread associated with it

pages/api/discord-thread.ts
```diff
@@ -1,7 +1,7 @@
 import { discordToken, evidenceRationalDiscordChannelId } from "constant";
 import { NextApiRequest, NextApiResponse } from "next";
 import {
-  DiscordThreadT,
+  VoteDiscussionT,
```
ryanwolhuter, line 4: renamed type as above

types/ui.ts
```diff
@@ -24,10 +22,6 @@ export type PanelTypeT =
   | "remind"
   | "delegation";
 
-export type VotePanelContentT = VoteT;
```
ryanwolhuter, line 27: when I initially added these types there were going to be more kinds of panel content. since there is only one now, these are redundant

types/voting.ts
```diff
@@ -297,15 +297,12 @@ export const DiscordMessageT = ss.object({
...
 
-export const DiscordThreadT = ss.object({
+export const VoteDiscussionT = ss.object({
   identifier: ss.string(),
   time: ss.number(),
   thread: ss.array(DiscordMessageT),
 });
-export type DiscordThreadT = ss.Infer<typeof DiscordThreadT>;
-
-export const DiscordThreadsT = ss.array(DiscordThreadT);
```
ryanwolhuter, line 307: removed these types because they are unused

components/Panel/VotePanel/VotePanel.tsx
```diff
@@ -56,11 +77,7 @@ export function VotePanel({ content }: Props) {
         isGovernance={isGovernance}
         voteNumber={resolvedPriceRequestIndex}
       />
-      {hasResults ? (
```
ryanwolhuter, line 59: this is the source of the bug where the discussion tab didn't show on active votes. active votes don't have results, so we'd never show the tabs

daywiss (approved):

daywiss (changes requested):
was having an issue with panel being empty
1. reload page
2. open more details in past request
3. see panel empty

ryanwolhuter:
> was having an issue with panel being empty
> 
> 1. reload page
> 2. open more details in past request
> 3. see panel empty

oof thanks for catching this, it doesn't happen on local. I think I know the cause though, will fix now

ryanwolhuter:
> was having an issue with panel being empty
> 
> 1. reload page
> 2. open more details in past request
> 3. see panel empty

@daywiss I think this takes the prize for the dumbest mistake of the year so far.

The tabs component takes a `defaultValue` which is the string title of a chosen tab. In the logic I did `hasResults ? "Results" : "Discussion"`. However, the title of the tab is "Result", not "Results" 🥲 

I have no idea why this worked on my local but not in production. I think it has something to do with the load order and the fact that the result tab is already the first tab in the list.

Anyway it is fixed now 👍

chrismaree:
I tested this out and it looks like it works great to me! thanks for adding this so quickly

chrismaree:
one strange thing is some of the messages have a link that is broken, like this:
2023-02-03 at 14 00 37@2x

if you click the link it takes you here https://media.discordapp.net/attachments/1058539248040542278/1058604844484669440/validation.log and is a dead link. 

when checking what the message actually has in discord it is a long "message log" https://discord.com/channels/718590743446290492/1059516345127030815/1059536132213317652 that looks like this in discord 
2023-02-03 at 14 02 25@2x

I looked into what the link that is produced from the discord serverless output and compared it to the one that is produced by the discord app itself:
- link produced in our dapp(dead link):  https://media.discordapp.net/attachments/1058539248040542278/1058604844484669440/validation.log
- link produced by discord (works to download the file): https://cdn.discordapp.com/attachments/1059516345127030815/1059536131856797746/validation.log

I played with this and basically what we need to do is update the discord serverless function such that if _any message contains a `media.discordapp.net`_ string we replace it with `cdn.discordapp.com`_. This will keep the formatting exactly the same as it is now but will actually let us pull down the log files if someone clicks the link. I think we should add this to this PR as well so we can test it all together. @ryanwolhuter can you do that?

daywiss (approved):
nice work

PR #182, opened 2023-02-07
feat: old designated voting contract accounts

ryanwolhuter:
### Motivation

there are a number of users who need to migrate their designated voting contracts to use the new delegation system. we should detect these users and tell them what to do.

I didn't want this to interfere with the existing error handling logic, so I made a variant of the error banner to show in this case instead, which is completely independent of the error state.

### Changes

**NOTE:** the large number of changes are because I updated storybook. I needed this upgrade so that I could use the new features that I'm using in the oracle dapp. **You can ignore the storybook related changes.**

* add data fetching logic for the `designated-voting` serverless function
* wire up this data to context
* show a variant of the error banner if this function detects the account

### Testing

This logic only runs when the connected account is one of those listed in `designated-voting.ts`. You can test this logic by taking one of those addresses (I used "0x718648C8c531F91b528A7757dD2bE813c3940608") and replacing the `address` passed in to `getIsOldDesignatedVotingAccount` in the `useIsOldDesignatedVotingAccount` hook.


.storybook/.babelrc
```diff
@@ -1,4 +0,0 @@
-{
```
ryanwolhuter, line 1: can ignore

.storybook/main.ts
```diff
@@ -1,18 +1,22 @@
-import type { StorybookConfig } from "@storybook/react/types";
+import type { StorybookConfig } from "@storybook/nextjs";
```
ryanwolhuter, line 1: can ignore

.storybook/preview.tsx
```diff
@@ -1,9 +1,7 @@
-import { addDecorator } from "@storybook/react";
+import { Decorator } from "@storybook/react";
```
ryanwolhuter, line 1: can ignore

components/ErrorBanner/ErrorBanner.tsx
```diff
@@ -1,10 +1,15 @@
-import { EthersErrorLink } from "components";
-import { mobileAndUnder } from "constant";
+import { EthersErrorLink } from "components/EthersErrorLink/EthersErrorLink";
```
ryanwolhuter, line 1: this component is the same, I just moved the styles to a separate file so I could re-use them in the new old designated voting warning banner

components/ErrorBanner/OldDesignatedVotingAccountWarningBanner.tsx
```diff
@@ -0,0 +1,72 @@
...
+  CloseButton,
+  CloseIcon,
+  ErrorMessage,
+  ErrorMessageWrapper,
+  IconWrapper,
+  WarningIcon,
+  Wrapper,
+} from "./styles";
+
+export function OldDesignatedVotingAccountWarningBanner() {
```
ryanwolhuter, line 15: component looks the same as the error banner, but it only deals with this particular situation

components/ErrorBanner/OldDesignatedVotingAccountWarningBanner.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const {
+    oldDesignatedVotingAccount: {
+      isOldDesignatedVotingAccount,
+      message,
+      designatedVotingContract,
+    },
+  } = useStakingContext();
+  const [closeClicked, setClosedClicked] = useLocalStorage(
+    "show-old-designated-voting-account-warning-banner",
```
ryanwolhuter, line 24: persist closing the banner in local storage so that it doesn't annoy the user

package.json
```diff
@@ -8,8 +8,8 @@
     "start": "next start",
     "lint": "next lint",
     "lint-fix": "next lint --fix",
-    "storybook": "start-storybook -p 6006",
-    "build-storybook": "build-storybook",
+    "storybook": "storybook dev -p 6006",
```
ryanwolhuter, line 11: changes were made automatically by storybook, can ignore

components/ErrorBanner/OldDesignatedVotingAccountWarningBanner.tsx
```diff
@@ -0,0 +1,79 @@
...
+  return (
+    <Wrapper>
+      <ErrorMessageWrapper>
+        <IconWrapper>
+          <WarningIcon />
+        </IconWrapper>
+        <ErrorMessage>
+          {message}:{" "}
+          <Link
+            href={`https://etherscan.io/contract/${designatedVotingContract}`}
```
daywiss, line 47: this doesnt work for some reason. sends me to a 404. also does this matter if its on goerli?

daywiss (commented):
one issue

components/ErrorBanner/OldDesignatedVotingAccountWarningBanner.tsx
ryanwolhuter: I used `contract` in the url but should have used `address`. Pushed a fix now. 

afaik this does not apply on goerli. @chrismaree is that correct?

daywiss (approved):

PR #183, opened 2023-02-20
improve: getPendingRequests and unstakeTime

ryanwolhuter:
### Summary

Updates to the contract change some of the methods we interact with.

* We now have a view method `getPendingRequests` that we can call instead of having to static call `getPendingRequestsPostUpdate`.

* Figuring out when a user can execute unstake has been simplified — now the `voterStakes` view function returns `unstakeTime` **which is the time that the user can execute the unstake.**

 I had to update both of these in this one PR because the code would not build with the out of date method calls. 
 
 ### Changes
 
 * update the call to `getPendingRequests`
 * remove `unstakeRequestTime` and use the new `unstakeRequestTime` field instead.


web3/queries/staking/getStakerDetails.ts
```diff
@@ -1,24 +1,16 @@
...
-  const unstakeCoolDown = await getUnstakeCoolDown(votingContract);
-  const { pendingUnstake, unstakeRequestTime, delegate, rewardsPaidPerToken } =
+  const { pendingUnstake, unstakeTime, delegate, rewardsPaidPerToken } =
     result ?? {};
-  const unstakeRequestTimeAsDate = new Date(Number(unstakeRequestTime) * 1000);
-  const canUnstakeTime = getCanUnstakeTime(
-    unstakeRequestTimeAsDate,
-    unstakeCoolDown.toNumber()
-  );
+  const canUnstakeTime = new Date(Number(unstakeTime) * 1000);
```
daywiss, line 10: this is not a frontend issue, but in practice this doesnt work too well becuase block time differs from clock time and in my testing this is really out of sync, leaving the user thinkign they can claim, when time is still pending. this might be ok though once we raise unstake time to over 1 minute.

daywiss (approved):
nice

PR #184, opened 2023-02-20
feat: fix commit/reveal for delegatee

daywiss:
## motivation
Voting contracts have changed their api with regard to commit reveal for delgatees, we want to fix that

## changes
this only changes the logic around committing as a delegatee. they now commit based on delegators address in the commit hash. this is enough to get the functions working correctly again.


ryanwolhuter (approved):
nice, much simpler

PR #186, opened 2023-02-23
fix: copy around connected wallet delegate

daywiss:
## motivation
copy was wrong when logged in as a delegate in wallet settings

## changes
corrrectly switch between copy given delegate/delegator setting

image
image


ryanwolhuter (approved):
👍

PR #187, opened 2023-02-23
fix: address link for delegate

daywiss:
## motivation
links to address were being incorrectly formatted inside panel when notifying of a pending delegate request

## changes
we did not have a function for configuring address links in the config, were incorrectly using the transaction link generator. this adds an address link generator for each chain.

### DIFF
--- a/components/pages/WalletSettings/PendingRequests.tsx
+++ b/components/pages/WalletSettings/PendingRequests.tsx
@@ -107,8 +107,9 @@ export function PendingRequests({
               <NextLink
                 href={config.makeTransactionHashLink(transactionHash)}
                 passHref
+                target="_blank"
               >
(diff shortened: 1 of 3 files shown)


components/pages/WalletSettings/PendingRequests.tsx
```diff
@@ -107,8 +107,9 @@ export function PendingRequests({
               <NextLink
                 href={config.makeTransactionHashLink(transactionHash)}
                 passHref
+                target="_blank"
               >
-                <A target="_blank">View Transaction</A>
```
daywiss, line 111: unrelated fix, but this was not opening in new tab

ryanwolhuter (approved):
nice

PR #195, opened 2023-02-27
fix: fix reward estimation

daywiss:
## motivation
when claiming rewards on a quickly accumulating account, we would see actual claim amounts inconsistent with the amount shown on UI. this would sometimes under or overestimate rewards by a large margin.

## changes
In order to calculate the estimated rewards in real time, we need to know the current state of rewards on chain at a specific timestamp, and we need the length of time which has elapsed since that was fetched.  we were not fetching the current block timestamp when querying the data, but rather using "date.now".  this could cause wildly different "fetch times" since block time and date.now can be way off.  as a result estimations could be off by a large amout.  We change this to fetch the block timestamp when we fetch the reward data. This seems to give more consistent estimations, after 5 claims, it underestimates every time, which is expected since we cant estimate when the claim transaction will be mined, and rewards will accrue until tx is mined.

image
image

image
image

### DIFF
--- a/hooks/queries/rewards/useRewardsCalculationInputs.ts
+++ b/hooks/queries/rewards/useRewardsCalculationInputs.ts
@@ -21,6 +21,7 @@ export function useRewardsCalculationInputs(addressOverride?: string) {
         rewardPerTokenStored: BigNumber.from(0),
         cumulativeStake: BigNumber.from(0),
         updateTime: BigNumber.from(Date.now()),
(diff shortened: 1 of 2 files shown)


smb2796:
Would like a review from @ryanwolhuter as well - but I did some quick testing and it does look like it improved the accuracy quite a bit! So looks good to me and thanks for fixing @daywiss !

ryanwolhuter (approved):
great work, well done for figuring it out so quickly

PR #197, opened 2023-02-27
fix: use yyyy mm dd date format

ryanwolhuter:
### Motivation

Use the date format that is most common in the world. To specify this format while still using an English-speaking locale, we use the en-CA (Canada) date locale 🍁 

### Changes

* Use en-CA locale when formatting dates

### DIFF
--- a/components/Panel/VotePanel/Discussion.tsx
+++ b/components/Panel/VotePanel/Discussion.tsx
@@ -1,6 +1,7 @@
 import { Button, LoadingSpinner } from "components";
 import { discordLink, mobileAndUnder, red500 } from "constant";
 import { format } from "date-fns";
+import { enCA } from "date-fns/locale";
 import { addOpacityToHsl } from "helpers";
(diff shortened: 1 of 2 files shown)


components/Panel/VotePanel/Discussion.tsx
```diff
@@ -1,6 +1,7 @@
 import { Button, LoadingSpinner } from "components";
 import { discordLink, mobileAndUnder, red500 } from "constant";
 import { format } from "date-fns";
+import { enCA } from "date-fns/locale";
```
daywiss, line 4: what is this, cant find it in date docs?

ryanwolhuter: https://date-fns.org/v2.29.3/docs/Locale
https://en.wikipedia.org/wiki/Date_format_by_country

daywiss (approved):
id just suggest commenting on this becuase its not obvious what and why we are using this setting

PR #198, opened 2023-02-27
feat: add warning when user changing prev vote from seperate account

daywiss:
## motivation
When your delegate/delegator commits, you could potentially change their vote on your account. we wanted to warn the user

## changes
This adds a hover info when this behavior is detected
image


ryanwolhuter (approved):

smb2796 (approved):
Copy looks good to me!

PR #200, opened 2023-02-27
feat: add notification that user is delegator

daywiss:
## motivation
We currently show the user they are delegate, but not delegator. We should show both

## changes
this just enables the same type of notification if the user is curently a delegator, identical in style to the delegate

image

### DIFF
--- a/components/Header/Header.tsx
+++ b/components/Header/Header.tsx
@@ -62,15 +62,18 @@ export function Header() {
   const status = getDelegationStatus();
   const showDelegationNotification =
     !getDelegationDataLoading() &&
-    (status === "delegate" || status === "delegate-pending");
+    (status === "delegate" ||
+      status === "delegate-pending" ||
+      status === "delegator");
   const showV1RewardsNotification = v1Rewards?.totalRewards.gt(0);
   const isDelegate = status === "delegate";
+  const isDelegator = status === "delegator";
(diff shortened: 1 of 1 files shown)


components/Header/Header.tsx
```diff
@@ -117,6 +120,13 @@ export function Header() {
                     <Link href="/wallet-settings">delegate wallet</Link>
                   </>
                 )}
+                {isDelegator && (
+                  <>
+                    <Dot />
+                    &nbsp;&nbsp;Connected as{" "}
```
ryanwolhuter, line 126: Using spaces like this is an antipattern that can make things get a bit weird.

Rather remove the `&nbsp;` and add margin to the `<Dot />` like so:

```
const Dot = styled.span`
  height: 8px;
  width: 8px;
  // change
  margin-right: 8px;
  // other suggestion
  background-color: var(--color);
  border-radius: 50%;
  display: inline-block;
`;
```

Note: I also suggested changing the background color to `var(--color)` so that it can inherit the color of the notification and therefore also be used in red notifications too with no other modifications.

ryanwolhuter (approved):

components/Header/Header.tsx
daywiss: alright ill give it a try

PR #201, opened 2023-02-28
fix: handle wrong chain

ryanwolhuter:
### Motivation

Having the wrong chain logic tied up with the general error handling logic was causing some issues. When you connect with the wrong chain, the app would flash the "wrong chain" message and then quickly remove it. The app would then show an "error fetching data" message instead. 

Since being connected to the wrong chain is a special situation that makes it so that no user dependent data has meaning and no actions can be taken, I've elected to move the logic for handling this situation out of the `ErrorContext`. Instead, we add a property to the `WalletContext` called `isWrongChain` which we can use to show/hide a special error banner. This error banner looks the same as the others but is different in that it does not have a "close" button — the only way to get rid of this error is to switch to the correct chain.

I've also added `!isWrongChain` to all of the queries which depend on the connected chain. I have not disabled the query for past votes and non-chain queries like contentful data and augmented requests as these still work when given past votes.

### Changes

* Add wrong chain error banner
* Move wrong chain logic to wallet context
* Erase provider and signer when is wrong chain
* Disable chain queries when is wrong chain

### DIFF
--- a/components/ErrorBanner/ErrorBanner.tsx
+++ b/components/ErrorBanner/ErrorBanner.tsx
@@ -1,5 +1,6 @@
 import { EthersErrorLink } from "components/EthersErrorLink/EthersErrorLink";
-import { useErrorContext } from "hooks";
+import { wrongChainMessage } from "constant";
+import { useErrorContext, useWalletContext } from "hooks";
 import { ErrorOriginT } from "types";
 import {
   CloseButton,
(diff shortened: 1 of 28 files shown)


components/ErrorBanner/ErrorBanner.tsx
```diff
@@ -12,27 +13,52 @@ import {
...
   const { errorMessages, removeErrorMessage } = useErrorContext(errorOrigin);
 
-  if (!errorMessages.length) return null;
+  if (!errorMessages.length && !isWrongChain) return null;
 
   return (
     <Wrapper>
-      {errorMessages.map((message) => (
-        <ErrorMessageWrapper key={message?.toString()}>
+      {isWrongChain ? (
```
ryanwolhuter, line 23: I'm treating being on the wrong chain as a special case here, since there's not much point saying "something went wrong fetching the data" when they're on the wrong chain and everything will obviously fail

components/ErrorBanner/ErrorBanner.tsx
```diff
@@ -12,27 +13,52 @@ import {
...
+        <Error message={wrongChainMessage} />
```
ryanwolhuter, line 24: No `removeErrorMessage` for this one, so the "close" button doesn't show. I think this makes sense — the only way to get rid of this error is to switch to the correct chain

daywiss (approved):
good find

PR #203, opened 2023-02-28
fix: unstake cooldown always shown as seconds

ryanwolhuter:
### Motivation

The date-fns library assumes that if you pass in a value to format as an interval with a unit like `{ seconds: value }` then you want the output to be in that unit, i.e. seconds in this case. Passing it in as an interval tells date-fns to format it with the least values instead like x days y hours z minutes etc.

### Changes

* Pass `unstakeCoolDown` to date-fns `format` as an interval


daywiss (approved):
nice trick

PR #205, opened 2023-02-28
fix: pagination showing 2 1s

ryanwolhuter:
### Motivation

The original pagination logic tried to account for fitting in buttons into a space where less than the default number would fit. This is actually pointless because that would only happen on a screen less than 200px wide, which is not a type of device that exists. This logic was overly complicated and caused a bug where the number of buttons and whether to show a "go to first page" button was calculated wrong. This PR removes the complicated logic and fixes the math mistake.

### Changes

* Remove dependency on screen width in number of buttons calculation
* use `>` instead of `>=`  for `hasMorePagesThanButtons`
* Account for `numberOfPages === defaultNumberOfButtons + 1` when calculating number of buttons

### DIFF
--- a/components/Pagination/Pagination.tsx
+++ b/components/Pagination/Pagination.tsx
@@ -1,10 +1,9 @@
 import { Dropdown } from "components";
-import { grey800, mobileAndUnder, white } from "constant";
+import { grey800, tabletAndUnder, white } from "constant";
 import { addOpacityToHsl } from "helpers";
-import { usePaginationContext, useWindowSize } from "hooks";
+import { usePaginationContext } from "hooks";
 import PreviousPage from "public/assets/icons/left-chevron.svg";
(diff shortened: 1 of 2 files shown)


daywiss (approved):
i noticed this yesterday but failed to report it! nice fix.
fyi not critical but this does have an issue where you go to the last page with 10 items showing, then switch to like 20 or 50 items, you wont have a valid page selected anymore.  somewhere shoudl be `currentPage = min(pages.length,currentpage)` or something

ryanwolhuter:
> i noticed this yesterday but failed to report it! nice fix. fyi not critical but this does have an issue where you go to the last page with 10 items showing, then switch to like 20 or 50 items, you wont have a valid page selected anymore. somewhere shoudl be `currentPage = min(pages.length,currentpage)` or something

Nice catch! noted and fixed

PR #213, opened 2023-03-02
feat: add apr override

daywiss:
## motivation
we want to override the apr of 0 to something more reasonable

## changes
this sets an env to override, remove env to reset back to dynamic calc.
image


smb2796 (approved):
LGTM

PR #214, opened 2023-03-02
feat: add independent signing button

daywiss:
## motivation
user is unable to sign when there is no active vote. this means they may not be able to see historical votes if they had never signed.

## changes
this adds a new button that works independently of the current vote. its a somewhat risky change since this logic is mixed in with old working logic. its been tested through commit reveal on goerli, as well as no votes on mainnet to check for correect behavior. 

image


daywiss:
closing this in favor of https://github.com/UMAprotocol/voter-dapp-v2/pull/218

PR #217, opened 2023-03-03
feat: address request augmentation issue with augmented cross-chain events

chrismaree:
- wip
- feat: address request augmentation issue with augmented cross-chain events


ryanwolhuter (approved):
👍

daywiss (approved):
working for me. i think we can eventually generalize this to other chains if we ever need to

PR #218, opened 2023-03-03
feat: use graph for revealing past votes

daywiss:
## motivation
We currently arent able to decode historical votes made from the v1 contract.

## changes
This switches the app to defer to the graphs reveal values when looking up past votes, but maintains the onchain decription for active votes. This automatically looks up desinated voting factory address, but we can override if needed with env var.  btw, i realize now theres a serverless function for this, could switch to that if desired. 

image

### DIFF
--- a/hooks/queries/user/useDesignatedVotingV1Address.ts
+++ b/hooks/queries/user/useDesignatedVotingV1Address.ts
@@ -1,0 +1,15 @@
+import { useQuery } from "@tanstack/react-query";
+import { getDesignatedVotingV1Address } from "helpers";
+
+export function useDesignatedVotingV1Address(address: string) {
+  return useQuery(
+    ["designatedVotingV1Address", address],
+    () => getDesignatedVotingV1Address(address),
+    {
+      enabled: !!address,
+      initialData: undefined,
+      onError: (err) =>
+        console.error("Error fetching designated voting v1 address", err),
+    }
+  );
+}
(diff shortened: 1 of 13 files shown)


smb2796:
Looks good! Approving, but might want to wait for code review from @chrismaree @ryanwolhuter before merging. Also looks like something is breaking in ci

smb2796 (approved):

ryanwolhuter:
@smb2796 looks like the build is only failing on netlify. it works on vercel and on my local. I actually don't have access to the team's netlify so I can't check what's going on. can you please give me access? also why do we even have this netlify deployment? I don't recall us ever using it?

hooks/queries/user/useDesignatedVotingV1Address.ts
```diff
@@ -0,0 +1,15 @@
+import { useQuery } from "@tanstack/react-query";
+import { getDesignatedVotingV1Address } from "helpers";
+
+export function useDesignatedVotingV1Address(address: string) {
+  return useQuery(
+    ["designatedVotingV1Address", address],
+    () => getDesignatedVotingV1Address(address),
+    {
+      enabled: !!address,
```
ryanwolhuter, line 9:
```suggestion
      enabled: !!address && !isWrongChain,
```

we should add an `isWrongChain` check so that this doesn't throw on the wrong chain

ryanwolhuter (approved):
works perfectly for me. I'll figure out why the build is failing on netlify when I get access. My current guess is something to do with env vars — I know I haven't set any in netlify since afaik we aren't using it

daywiss:
@ryanwolhuter it will fix itself most likely when i push a change

PR #222, opened 2023-03-07
fix: make signature match old dapp by default

mrice32:
This change aligns the signature process with the old UMA voter dapp so the derived keys match that dapp. In the future, it could make sense to establish a message "versioning" system that determines which message to use based on which round is being used. However, for now, it makes sense to just align the two messages to start.


mrice32:
@ryanwolhuter @daywiss please do not merge this. We should ensure that:
1. We are not in the middle of an active voting round.
2. We have a way to ensure the signatures get updated.

We can discuss this later this week!

PR #225, opened 2023-03-07
feat: make phase length configurable

daywiss:
## motivation
phase length differs by primary chain ( goerli shorter than mainnet), this should be configurable

## change
this adds an env to allow phase length to be set in seconds NEXT_PUBLIC_PHASE_LENGTH.  if not specified it wil default to mainnet length.  env has been updated for goerli at a value of 600


ryanwolhuter (approved):
I knew we would miss something like this with the launch. glad it wasn't too hectic

PR #226, opened 2023-03-08
fix: stop showing incorrect decrypt errors

daywiss:
## motivation
people are seeing decrypt error on active votes

## changes
this is happenign because we were not correctly removing v1 encrypted past votes, causing decrypt error during our decryption query


smb2796 (approved):

mrice32 (approved):
LGTM!

PR #227, opened 2023-03-08
feat: refac vote page to show only 5 past votes

daywiss:
## motivation
we want to show a limited view of past votes, up to 5. currently we show the entire past votes table on the main page.

## changes
this refactors the vote component to separate the logic of "active", "upcoming" and "past" views into their own functions. the "past" view is re-used in both active and upcoming phases, so we have less duplicate code. 

upcoming view
vote-upcoming

active view
active-vote

past view
image

### DIFF
--- a/components/Votes/Votes.tsx
+++ b/components/Votes/Votes.tsx
@@ -304,62 +304,98 @@ export function Votes() {
         return getActiveVotes();
       case "upcoming":
         return getUpcomingVotes();
-      case "past":
-        return getPastVotes();
+      default:
+        return [];
     }
   }
 
-  function determineTitle() {
-    const status = getActivityStatus();
-    switch (status) {
-      case "active":
-        return "Active votes:";
-      case "upcoming":
-        return "Upcoming votes:";
-      default:
-        return "Past votes:";
-    }
+  const activityStatus = getActivityStatus();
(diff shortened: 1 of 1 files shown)


components/Votes/Votes.tsx
```diff
@@ -304,62 +304,98 @@ export function Votes() {
         return getActiveVotes();
       case "upcoming":
         return getUpcomingVotes();
-      case "past":
```
daywiss, line 307: this is no longer used anywhere

components/Votes/Votes.tsx
```diff
@@ -405,51 +441,74 @@ export function Votes() {
...
+        {pv ? (
+          <>
+            <Divider />
+            {pv}
+          </>
+        ) : null}
+      </>
+    );
+  }
+  function upcomingView() {
```
ryanwolhuter, line 453: not important but interesting — these don't actually have to be functions if they aren't taking props. you can just do `const upcomingView = (<>jsx stuff<>)` and put it in the other jsx the same as you did here like `{upcomingView}`

daywiss: probably true, i am composing one of them though, but the others probably dont have to be

PR #229, opened 2023-03-09
 fix: use non wallet provider for reading data 

daywiss:
## motivation
we are seeing some weird behavior when fetching data from providers. it turns out we are using the metmask provider once we connect wallet to pull data.

## changes
previouslly we had a single instance of a voting and voting token contract which was tied to the wallets provider and signer.  now we have 2 instances, one for reading, using our own provider, and one for writing using the wallets signer. this should make quering more predictable and have less errors, hopefully improving usability for people who are having connectivity issues.

### DIFF
--- a/components/Panel/ClaimPanel.tsx
+++ b/components/Panel/ClaimPanel.tsx
@@ -22,7 +22,7 @@ import {
 const minimumAmountClaimable = parseEtherSafe(".01");
 
 export function ClaimPanel() {
-  const { voting } = useContractsContext();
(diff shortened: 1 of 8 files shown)


ryanwolhuter (approved):
makes sense to me. I just voted on this preview link and it worked 👍 

I think we should refactor this to have names like `votingReader` too in the future but that will be a bigger task

mrice32 (approved):
I was initially hoping we could just take the signer they gave us and attach our own provider to it, but I think what you have here is the best solution!

PR #230, opened 2023-03-13
fix: outstanding rewards

md0x:
Changes proposed in this PR:

- Fixed outstanding rewards off-chain calculation


chrismaree (approved):
I've tested this in prod and it looks to work correctly now!

daywiss (approved):
thanks!

PR #231, opened 2023-03-14
fix: do not parses your vote as number

daywiss:
## motivation
we are seeing magic number shown in a truncated form, this makes it difficult to verify we committed the right value

## changes
this removes a line that parses the value as  a number, and just passes through the raw decoded string
image

### DIFF
--- a/components/VotesList/VotesListItem.tsx
+++ b/components/VotesList/VotesListItem.tsx
@@ -130,12 +130,7 @@ export function VotesListItem({
   }
 
   function getDecryptedVoteAsString() {
-    return getDecryptedVoteAsNumber()?.toString();
-  }
-
-  function getDecryptedVoteAsNumber() {
-    const formattedString = getDecryptedVoteAsFormattedString();
-    return formattedString ? Number(formattedString) : undefined;
+    return getDecryptedVoteAsFormattedString();
   }
 
   function showVoteInput() {


components/VotesList/VotesListItem.tsx
```diff
@@ -130,12 +130,7 @@ export function VotesListItem({
   }
 
   function getDecryptedVoteAsString() {
-    return getDecryptedVoteAsNumber()?.toString();
-  }
-
-  function getDecryptedVoteAsNumber() {
-    const formattedString = getDecryptedVoteAsFormattedString();
-    return formattedString ? Number(formattedString) : undefined;
```
daywiss, line 138: @ryanwolhuter not sure what the reason was we were doign this.  will this break anything removing this cast?

ryanwolhuter: We introduced this code ages ago in [this pr](https://github.com/UMAprotocol/voter-dapp-v2/pull/28/files#diff-6efe6a4065c26b194db19e01be71cf77b89070bb27cfb9434683505bb24f7b7f)

I think at the time there was an issue with the type expected by the dropdown and the type expected by the input. The problem fixed in that PR was that using type number in the input causes it to put a "," in the input text, so we switched to string. I think I just left the cast in because I thought we might still need to use the number at some other point.

It doesn't seem like there's any issue with the input when you remove the cast, and the existing input doesn't expect a number. I think you can remove it without breaking stuff.

PR #233, opened 2023-03-16
feat: add dynamic polymarket dropdown parsing

daywiss:
## motivation
people are having a hard time selecting correct options when selecting polymarket votes, and we arent able to detect the options

## changes
this uses a regex to match the p1,p2,p3 values, which only works in the case where there are those 3 options, and builds up options list based on matches. it will also always append magic number ( too early ) and a custom input

image

This will only display if its able to find at least 3 options.

### DIFF
--- a/helpers/voting/polymarket.ts
+++ b/helpers/voting/polymarket.ts
@@ -1,5 +1,6 @@
 import { earlyRequestMagicNumber } from "constant";
 import { DropdownItemT } from "types";
+import chunk from "lodash/chunk";
 
 export function checkIfIsPolymarket(
   decodedIdentifier: string,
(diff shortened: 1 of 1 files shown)


chrismaree:
How can I test this with a wide set of inputs? This looks awesome!

daywiss:
@chrismaree you can scroll through past requests and look at the options in details 
image

helpers/voting/polymarket.ts
```diff
@@ -14,6 +15,46 @@ export function checkIfIsPolymarket(
 
   return isPolymarket;
 }
+
+// this will only work when there are exactly 3 or more options, which should match most polymarket requests
+// it will only parse 3 options, omitting p4, which is assumed to be "too early".
+function dynamicPolymarketOptions(
+  decodedAncillaryData: string
+): DropdownItemT[] {
+  const resData = decodedAncillaryData.match(
```
ryanwolhuter, line 24: maybe we should handle the situation where they forget the spaces between the p things

ryanwolhuter: I'm sure you could do it with the regex but I'd rather jump off a cliff than figure that one out so yeah

daywiss: i actually was using replace instaed of replaceAll. i can get regex to work with no spaces, but i think this will cause other complications, it will break the formatting of the labels if they are mutliple words, so ill leave it for now

ryanwolhuter (approved):

chrismaree:
one thing that would be nice is to update the donut chart to contain all options. for example consider this request with 5 options.

2023-03-16 at 16 52 02@2x

in this, everyone voted for the same outcome:
2023-03-16 at 16 52 37@2x


we did not show these in the donut chart as a result. would it be better to show all 5 and show them at 0?


also, unrelated to the above point, but it still really triggers me how we truncate on the donut chart, like this:
2023-03-16 at 16 53 06@2x

daywiss:
> One question I have is this will ONLY apply to polymarket requests, right? we are using the function dynamicPolymarketOptions and this requires the identifier and runs a check on ancillary data as well. this is correct, right?

@chrismaree yes there is higher level logic that determines if this is a polymarket request before calling this parsing 

> we did not show these in the donut chart as a result. would it be better to show all 5 and show them at 0?
> 
> also, unrelated to the above point, but it still really triggers me how we truncate on the donut chart, like this:
potentially, make some feature requests so @smb2796 can 👍 or 👎

smb2796:
This looks really good. I tested pretty extensively on goerli and mainnet, and everything lgtm. 

@chrismaree on the pie charts, agreed with @daywiss - lets have these be future feature requests unrelated to this PR. We can discuss in linear.

smb2796 (approved):

PR #234, opened 2023-03-16
feat: show upcoming votes when active vote

daywiss:
## motivation
we currently do not show an upcoming vote if there is an active vote

## changes
this simple feature was very risky to implement with current code base. this needed to be refactored to isolate past, upcoming, and active components, primarily to avoid breaking the delicate logic during an active vote.  the state between these components were previously shared with assumptions that we would only ever have active votes or upcoming votes independently.

This pr creates a new component for each table, and in the higher level page, composes them based on 4 different potential states: no active or upcoming votes, only an active vote, only upcoming vote, active and upcoming vote and no votes (active,upcoming,past). 

Things should look the same in the base case (active votes, with nothing upcoming, or just a single upcoming vote):
active-vote
revealed
image

now active + upcoming 
image

### DIFF
--- a/components/Votes/ActiveVotes.tsx
+++ b/components/Votes/ActiveVotes.tsx
@@ -1,0 +1,440 @@
+import { useConnectWallet } from "@web3-onboard/react";
+import {
+  Button,
+  IconWrapper,
+  Pagination,
+  Tooltip,
+  VotesList,
+  VotesListItem,
+  VotesTableHeadings,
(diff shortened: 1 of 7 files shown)


smb2796:
@daywiss it seems like this is high risk / low impact enough that we should absolutely wait until the current vote is done before merging. What do you think?

daywiss:
closed in favor of #240

PR #235, opened 2023-03-17
fix: merge revealed queries with existing cache

daywiss:
## motivation
theres an intermittent problem people are reporting, where they reveal but the app doesnt recognize they reveal.

## changes
unfortunately not able to reliably reproduce this problem, but it may be due to out of sync providers. when we write the reveal, we use the wallets signer, but when we read if we have revealed we use an infura provider, its possible these are sometimes out of sync.  this will cause problems with react query overwriting an empty revealed query after we have updated that cache locally when we know our tx revealed.  this change tries to merge the revealed table when a query to web3 is done to not blow away any locally modified values.

### DIFF
--- a/hooks/queries/votes/useRevealedVotes.ts
+++ b/hooks/queries/votes/useRevealedVotes.ts
@@ -1,4 +1,5 @@
-import { useQuery } from "@tanstack/react-query";
+import { useQuery, useQueryClient } from "@tanstack/react-query";
+import { VoteExistsByKeyT } from "types";
 import { revealedVotesKey } from "constant";
 import {
   useAccountDetails,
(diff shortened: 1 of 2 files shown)


hooks/queries/votes/useRevealedVotes.ts
```diff
@@ -19,7 +21,25 @@ export function useRevealedVotes(addressOverride?: string) {
 
   const queryResult = useQuery(
```
ryanwolhuter, line 22: makes sense!

this will achieve the same result in a bit of a cleaner way by using the built-in `onSuccess` hook

```suggestion
    const queryResult = useQuery({
    queryKey,
    queryFn: () => getRevealedVotes(voting, address, roundId),
    onSuccess(data) {
      queryClient.setQueryData<typeof data>(queryKey, (oldRevealedVotes) =>
        oldRevealedVotes ? { ...oldRevealedVotes, ...data } : data
      );
    },
    onError,
    initialData: {},
    enabled: !!address && !isWrongChain,
  });
```

daywiss: nice, will change this

ryanwolhuter (approved):

PR #236, opened 2023-03-20
improve: port oracle dapp pagination

ryanwolhuter:
### Motivation

The pagination logic in the new oracle dapp is much simpler and better than what was done in the voter dapp. Porting that logic here will simplify several upcoming tasks, and decrease the opportunities for bugs.

One thing that needed to be changed in the dapp logic is the way we get lists of votes from the VotesContext. I had originally implemented getter functions for handling this, but they are actually not necessary and caused rendering dependency issues with the new pagination component. I should have removed these long ago.

### Changes

* Port new oracle dapp pagination component
* Just return arrays of votes from the `VotesContext` instead of using redundant getter functions
* Rename vote lists to `ByKey` and `List` for clarity


daywiss (commented):
mostly works, but seeing an issue:
1. go to 50 per page. go to last page, switch back to 10 per page and the table dissapears. seems to happen when going from large number show down to smaller number.

daywiss:
@ryanwolhuter this is really weird, but im not seeing past votes show up at all. no errors in console, not sure if its related to this pr

daywiss (approved):
works great thanks!

PR #238, opened 2023-03-24
feat: add exit button to custom input

ryanwolhuter:
### Motivation

We need to be able to exit out of custom input for votes. Right now if you choose custom input, you're locked in.

**Note:** I've also been updating the relevant storybook code as I go along because it makes it easier for me to see what I'm doing. You don't need to review the story code.

### Changes

* Add exit button to text input
* Small css fixes


components/ErrorBanner/OldDesignatedVotingAccountWarningBanner.tsx
```diff
@@ -71,7 +71,7 @@ const Link = styled(NextLink)`
   font: inherit;
   color: inherit;
   text-decoration: underline;
-  transition: opacity var(--animation-duration);
```
ryanwolhuter, line 74: this css var was copied over from somewhere else, but it's not defined in this app.

.storybook/preview.tsx
```diff
@@ -37,16 +36,14 @@ export const decorators: Decorator[] = [
...
-                    <DelegationProvider>
-                      <VotesProvider>
-                        <PanelProvider>
-                          <GlobalStyle />
-                          <Story />
-                        </PanelProvider>
-                      </VotesProvider>
-                    </DelegationProvider>
-                  </PaginationProvider>
+                  <DelegationProvider>
```
ryanwolhuter, line 39: we removed the pagination provider in a previous pr

components/Notifications/Notification.tsx
```diff
@@ -78,12 +78,7 @@ const IconWrapper = styled.div`
   height: 32px;
 `;
 
-const SuccessIcon = styled(Success)`
-  path {
-    fill: var(--white);
-    stroke: var(--green);
-  }
-`;
+const SuccessIcon = styled(Success)``;
```
ryanwolhuter, line 81: I think these are old styles. the icon looks right as is without them, and having them cuts off the edges of the circle

components/VotesList/VotesListItem.tsx
```diff
@@ -83,23 +83,22 @@ export function VotesListItem({
   const isRolled = rollCount > 0;
   const wrapperRef = useRef<HTMLTableRowElement>(null);
   const existingVote = getDecryptedVoteAsFormattedString();
-  useEffect(() => {
-    if (!options) return;
```
ryanwolhuter, line 87: I think this is from old code — there's no need to do this check. and having an object like this in the dependency array causes this hook to run on every render, which means if you try to change the input back from custom it will get changed back to custom again on every render.

components/VotesList/VotesListItem.tsx
```diff
@@ -83,23 +83,22 @@ export function VotesListItem({
...
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [options, decryptedVote]);
+  }, [decryptedVote]);
 
   useEffect(() => {
     if (wrapperRef.current) {
       setWrapperWidth(wrapperRef.current.offsetWidth);
     }
-  }, [wrapperRef.current?.offsetWidth]);
+  }, [width]);
```
ryanwolhuter, line 101: this actually never worked. turns out refs never cause rerenders, so having a ref in the dependency array like this does nothing. it looked like it worked because other things would cause a rerender. using the width as the dependency solves this bug

components/VotesList/VotesListItem.tsx
```diff
@@ -55,7 +56,6 @@ export function VotesListItem({
   isDirty = false,
   delegationStatus,
 }: Props) {
-  const { signer } = useWalletContext();
```
ryanwolhuter, line 58: see below comment about disabling text input

components/VotesList/VotesListItem.tsx
```diff
@@ -380,13 +383,20 @@ export function VotesListItem({
               onSelect={onSelectVote}
             />
           ) : (
-            <TextInput
-              value={selectedVote ?? getDecryptedVoteAsString() ?? ""}
-              onInput={selectVote}
-              maxDecimals={maxDecimals}
-              type="number"
-              disabled={!signer}
```
ryanwolhuter, line 388: I think disabling the text input when there's no signer is from old code as well. We don't disable the dropdown, we just disable the commit button.

daywiss (commented):
works great for me, but i think from a UX perspective something is wrong;
currently hitting X only changes back to dropdown, without clearning your custom input. i think this is misleading because typically this means "clear input".  I think the solution here is to actually clear users input AND switch back to dropdown. since goign back to dropdown we can no longer see our custom input, it seems a bit weird to me.

ryanwolhuter:
@daywiss I've updated the component to clear the input when the user clicks close. 

I've prepared stories for each of the situations relevant here:

[No existing vote](https://62c6c91636b0d76a1ed19f92-gupzppagwk.chromatic.com/?path=/story/pages-vote-page-voteslistitem--polymarket-not-committed)

[Existing vote is normal option](https://62c6c91636b0d76a1ed19f92-gupzppagwk.chromatic.com/?path=/story/pages-vote-page-voteslistitem--polymarket-committed)

[Existing vote is custom input](https://62c6c91636b0d76a1ed19f92-gupzppagwk.chromatic.com/?path=/story/pages-vote-page-voteslistitem--polymarket-committed-custom-value)

daywiss (approved):
great!

PR #240, opened 2023-03-28
improve: refactor vote page

ryanwolhuter:
### Motivation

The `Votes` component has become too complex, because it is responsible for rendering several permutations of past votes, upcoming votes, and active votes.

I've split out these views into separate components. Not much logic has changed — all of the business logic (which was only relevant to active votes) has just moved to the new `ActiveVotes` component. 

[I've created stories for each possible permutation here.](https://62c6c91636b0d76a1ed19f92-kmxwuzyvzl.chromatic.com/?path=/story/pages-vote-page-votepage)

### Changes

* Separate past, upcoming and active votes views
* Move voting business logic to active votes view
* Show votes on the page with the order active -> upcoming -> past


components/Input/TextInput.tsx
```diff
@@ -46,11 +49,17 @@ export function TextInput({
         maxLength={79}
         spellCheck="false"
       />
+      {!!onClear && (
```
ryanwolhuter, line 52: I moved the clear input button into the component so that we can use it in other places too

components/Votes/ActiveVotes.tsx
```diff
@@ -0,0 +1,396 @@
+import { useConnectWallet } from "@web3-onboard/react";
```
ryanwolhuter, line 1: all of the voting business logic has been moved here

components/Votes/ActiveVotes.tsx
```diff
@@ -0,0 +1,396 @@
...
+  const { votingWriter } = useContractsContext();
+  const { stakedBalance } = useStakingContext();
+  const { getDelegationStatus, getDelegatorAddress } = useDelegationContext();
+  const { openPanel } = usePanelContext();
+  const [{ connecting: isConnectingWallet }, connect] = useConnectWallet();
+  const { commitVotesMutation, isCommittingVotes } = useCommitVotes();
+  const { revealVotesMutation, isRevealingVotes } = useRevealVotes();
+  const [selectedVotes, setSelectedVotes] = useState<SelectedVotesByKeyT>({});
+  const [dirtyInputs, setDirtyInput] = useState<boolean[]>([]);
+  const [votesToShow, setVotesToShow] = useState(activeVotesList);
```
ryanwolhuter, line 58: the `activeVotesList` is now the only list of votes that this component ever deals with

components/Votes/ActiveVotes.tsx
```diff
@@ -0,0 +1,396 @@
...
+      canReveal: false,
+    };
+
+    const isCommit = phase === "commit";
+    const isReveal = phase === "reveal";
+    const hasStaked = stakedBalance?.gt(0) ?? false;
+    const hasSigner = !!signer;
+
+    const hasPreviouslyCommittedAll =
+      activeVotesList.filter((vote) => vote.decryptedVote).length ===
```
ryanwolhuter, line 103: Previously we were iterating over the `votesToShow` variable in this logic. This actually would have caused a bug if there were enough active votes to show the pagination. It would have ignored all the votes besides the current page.

components/Votes/ActiveVotes.tsx
```diff
@@ -0,0 +1,396 @@
...
+    if (!actionStatus.canReveal || !votingWriter) return;
+
+    revealVotesMutation({
+      voting: votingWriter,
+      votesToReveal: getVotesToReveal(),
+    });
+  }
+
+  function getVotesToReveal() {
+    return activeVotesList.filter(
```
ryanwolhuter, line 284: same for reveal

components/VotesList/VotesListItem.tsx
```diff
@@ -28,16 +28,15 @@ import Polymarket from "public/assets/icons/polymarket.svg";
...
 import { CSSProperties, ReactNode, useEffect, useRef, useState } from "react";
 import styled from "styled-components";
 import { ActivityStatusT, DropdownItemT, VotePhaseT, VoteT } from "types";
 export interface Props {
   vote: VoteT;
   phase: VotePhaseT;
-  selectedVote: string | undefined;
-  selectVote: (value: string | undefined) => void;
-  clearVote: () => void;
+  selectedVote?: string | undefined;
```
ryanwolhuter, line 37: made these optional so the past and upcoming views can leave them out

pages/index.tsx
```diff
@@ -1,8 +1,34 @@
...
+  Votes,
+} from "components";
+import { Strong } from "components/pages/styles";
+import { truncateDecimals } from "helpers";
+import { useGlobals } from "hooks";
 import type { NextPage } from "next";
 
 const VotePage: NextPage = () => {
-  return <Vote />;
+  const {
```
ryanwolhuter, line 15: Having this redundant component within a component structure is unnecessary. I will update the other pages to work like this in another PR.

components/Votes/Votes.tsx
```diff
@@ -1,570 +1,29 @@
...
-    return activeView();
-  } else if (activityStatus === "upcoming") {
-    return upcomingView();
-  } else {
-    return pastView();
-  }
+  const { activeVotesList, pastVotesList, upcomingVotesList } =
+    useVotesContext();
+
+  const hasActiveVotes = activeVotesList.length > 0;
```
ryanwolhuter, line 12: This structure actually works out quite conveniently.

Active votes always go on the top. 

If we have active and upcoming votes, the upcoming votes go under the active votes.

If we have upcoming with no active, the upcoming will then be at the top because the active won't show.

Past votes always go to the bottom. If there are no active or upcoming, the past votes will then be the only list shown.

PR #241, opened 2023-04-12
feat: add hard coded assertion vote

daywiss:
## motivation
we wanted a quick fix for an upcoming vote coming from oracle v3

## changes
this hard codes the description for the assertion, it assumes we are not going to get anymore v3 disputes, and that we will generalize this soon

image

### DIFF
--- a/helpers/voting/getVoteMetaData.ts
+++ b/helpers/voting/getVoteMetaData.ts
@@ -16,6 +16,27 @@ export function getVoteMetaData(
   decodedAncillaryData: string,
   umipDataFromContentful: ContentfulDataT | undefined
 ): VoteMetaDataT {
+  // we are hard coding this because we have an upcoming vote, but the voterdapp cannot currently decode it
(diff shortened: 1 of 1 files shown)


helpers/voting/getVoteMetaData.ts
```diff
@@ -16,6 +16,27 @@ export function getVoteMetaData(
...
+    return {
+      title: decodedIdentifier,
+      description: "Asserted truth: 1+1=3",
+      umipOrUppLink: {
+        label: "umip-170".toUpperCase(),
+        href: "https://github.com/UMAprotocol/UMIPs/blob/7e4eadb309c8e38d540bdf6f39cee81a3e48d260/UMIPs/umip-170.md",
+      },
+      umipOrUppNumber: "umip-170",
+      options: [
+        { label: "True", value: "1" },
```
chrismaree, line 32: this will vote with the value `1e18` not `1`, right?

PR #244, opened 2023-04-20
fix: x to clear showing up on non dropdown

ryanwolhuter:
Don't show the "x" button on the input when it's not a custom input. I initially did this on purpose because I thought it could be useful, but I understand now that it can be confusing for the user.

**I also upgraded storybook.** This was done automatically and the changes can be ignored. **I've made a comment by the one line of logic that was changed to solve this bug.**

**[Story](https://62c6c91636b0d76a1ed19f92-fjbqmrfhfv.chromatic.com/?path=/story/pages-vote-page-votepage--active-commit-with-pagination)**


components/VotesList/VotesListItem.tsx
```diff
@@ -389,7 +389,7 @@ export function VotesListItem({
             <TextInput
               value={selectedVote ?? getDecryptedVoteAsString() ?? ""}
               onInput={selectVote}
-              onClear={exitCustomInput}
+              onClear={isCustomInput ? exitCustomInput : undefined}
```
ryanwolhuter, line 392: Only provide onClear when we are dealing with a custom input.

daywiss: in the story book link its showing an x on some inputs without a dropdown 
image

daywiss (approved):

PR #245, opened 2023-04-20
fix: set rewards to 0 on claim and withdraw

daywiss:
## motivation
it seems like sometimes especially on mobile, when we claim and stake the amount to claim doesnt zero out as expected

## changes
difficult to repro on desktop, but it looks like we were trying to reset outstanding rewards on a defunct key, so updated the reset to the correct rq key.


ryanwolhuter (approved):
nice catch

PR #246, opened 2023-04-24
feat: remove error clearing from rq queries

daywiss:
## motivation
errors are flashing, unable to be read when on main page, such as cancelling a tx

## changes
this removes the clear errors from react query queries, these tend to run frequently, particularly when re-focusing on page.  this means error banner sticks around until user manually closes it. 

future work - may need to have some kind of global error manager to clear errors after some seconds


ryanwolhuter (approved):
👍

PR #247, opened 2023-04-25
fix: remove non deterministic values from mocks

ryanwolhuter:
We have some story code that uses `Math.random` and `new Date`. These are different on every push to chromatic, creating false positives in the visual regression tests.

Remove these usages in favor of deterministic values.

### DIFF
--- a/stories/Pages/PastVotesPage.stories.tsx
+++ b/stories/Pages/PastVotesPage.stories.tsx
@@ -1,6 +1,7 @@
 import { Meta, Story } from "@storybook/react";
-import { defaultVotesContextState, VotesContext } from "contexts";
+import { VotesContext, defaultVotesContextState } from "contexts";
 import { BigNumber } from "ethers";
+import { uniqueId } from "lodash";
 import PastVotesPage from "pages/past-votes";
 import { VoteT } from "types";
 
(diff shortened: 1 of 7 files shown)


stories/Pages/PastVotesPage.stories.tsx
```diff
@@ -17,17 +18,13 @@ const mockPastVotes = Array.from({ length: 200 }).map((_, i) => ({
   decodedAncillaryData:
     'q:"Had the following insured event occurred as of request timestamp: Tom ate Jerry?",ooRequester:0438e738cb7b1eb98b8cb9f26d6a11d59506b9dd',
   correctVote: 0,
-  uniqueKey:
-    "YES_OR_NO_QUERY-1666025124-0x713a224861642074686520666f6c6c6f77696e6720696e7375726564206576656e74206f63637572726564206173206f6620726571756573742074696d657374616d703a20546f6d20617465204a657272793f222c6f6f5265717565737465723a30343338653733386362376231656239386238636239663236643661313164353935303662396464" +
-    Math.random().toString(),
+  uniqueKey: uniqueId(),
```
daywiss, line 21: good idea

daywiss (approved):

PR #248, opened 2023-04-25
fix: show signing warning in table

daywiss:
## motivation
when we dont have a signature on file, we assume if we see a vote on the address, that its the delegator. but we should prefer to show user needs signature.

## changes
this checks if you have signature on file before continuing with rest of logic
image


ryanwolhuter (approved):
Nice catch

PR #249, opened 2023-04-25
improve: add mobile breakpoints to visual tests

ryanwolhuter:
By adding a list of breakpoints in our `preview` file for storybook, we tell chromatic to do snapshot tests of each of the stories at these breakpoints. This is great to catch regressions on smaller devices. We are already doing this in the oracle dapp.


daywiss (approved):
nice

PR #252, opened 2023-05-01
feat: add assertion details for oracle v3 requests

daywiss:
## motivation 
we want to be able to see oracle v3 claim data


helpers/voting/makePriceRequestsByKey.ts
```diff
@@ -74,6 +74,29 @@ function formatPriceRequest(
     : priceRequest.isGovernance;
   const rollCount = priceRequest.rollCount || 0;
   const resolvedPriceRequestIndex = priceRequest.resolvedPriceRequestIndex;
+  const isAssertion = ["assertionId:", "ooAsserter:", "childChainId:"].every(
```
daywiss, line 77: adds assertion meta data to vote type, could probably add the osnap stuff as well

components/Panel/VotePanel/Details.tsx
```diff
@@ -39,18 +41,25 @@ export function Details({
...
   const [showDecodedAdminTransactions, setShowDecodedAdminTransactions] =
     useState(false);
   const [showRawAncillaryData, setShowRawAncillaryData] = useState(false);
+  const [showRawClaimData, setShowRawClaimData] = useState(false);
 
   function toggleShowDecodedAdminTransactions() {
     setShowDecodedAdminTransactions(!showDecodedAdminTransactions);
   }
 
+  const { data: claim } = useAssertionClaim(assertionChildChainId, assertionId);
```
daywiss, line 56: just fetch individual claim data, this is indexed by assertion id, so its pretty fast, and also it will now cache by assertion and chain

ryanwolhuter:
@daywiss it's gonna be a pain to remove the osnap stuff. As it is, the logic never runs because we don't actually use it in the `getVoteMetaData` function. How do you feel about just merging that code so long (since it actually works) without activating it, and then opening a new PR to finish implementing it?

daywiss:
> @daywiss it's gonna be a pain to remove the osnap stuff. As it is, the logic never runs because we don't actually use it in the getVoteMetaData function. How do you feel about just merging that code so long (since it actually works) without activating it, and then opening a new PR to finish implementing it?

thats fine!

components/Panel/VotePanel/VotePanel.tsx
```diff
@@ -32,13 +35,19 @@ export function VotePanel({ content }: Props) {
     }
   );
 
+  const { data: claim } = useAssertionClaim(assertionChildChainId, assertionId);
+
+  const titleToShow = claim
```
daywiss, line 40: this is going to need to be truncated or something. claims could be really long

web3/contracts/createOOV3ContractInstances.ts
```diff
@@ -0,0 +1,77 @@
...
+    };
+  });
+  return instances;
+}
+
+export const instances = createOOV3ContractInstances();
+
+export function getInstance(desiredChainId: number) {
+  const found = instances.find(({ chainId }) => chainId === desiredChainId);
+  if (!found)
```
daywiss, line 74: this should also throw if `found.instance` is undefined, since providers are optional for each chain

helpers/util/misc.ts
```diff
@@ -22,3 +22,42 @@ export function makeTransactionHashLink(
...
+    case 43114:
+      return "https://snowtrace.io";
+    case 42161:
+      return "https://arbiscan.io";
+    case 80001:
+      return "https://mumbai.polygonscan.com";
+  }
+}
+
+export function makeBlockExplorerLink(
```
daywiss, line 53: nice util

ryanwolhuter (approved):

PR #254, opened 2023-05-02
improve: allow howitworks values to show as they load

daywiss:
## motivation
how it works values can take a long time to load

## changes
we we only displaying any values once all values were available, this would block showing values which may have loaded already, but were blocked from showing until everything loaded.  this changes it so that the values are displayed only when their direct depenencies are loaded.


ryanwolhuter (approved):
yeah at the time I believed that it looks better to show everything all at once, but experience has proved that isn't true

PR #257, opened 2023-05-04
feat: add vote stats during reveal

daywiss:
## motivation
we want to see vote information during reveal

## changes
in order to do this, we need to query the graph for active votes.  this will attach the "results" and "participation" data needed for each active request, and this data will automatically populate the results panel.  theres a known issue, that if the vote rolls the results and participation will not clear, but we dont experience rolls on mainnet, so its not a critical bug.  the graph will need to be updated to resolve this. 

image


ryanwolhuter (approved):
nice

PR #258, opened 2023-05-05
improve: remove redundant context getters and memoize the values

ryanwolhuter:
### Motivation

When we use an object as the `value` prop of a react context, that object is recreated on every render. This has performance implications. From the [react docs](https://react.dev/reference/react/useContext#optimizing-re-renders-when-passing-objects-and-functions):

"Whenever MyApp re-renders (for example, on a route update), this will be a different object pointing at a different function, so React will also have to re-render all components deep in the tree that call useContext(AuthContext)."

Here's a codesandbox showing this problem:

https://codesandbox.io/s/jj611c?file=/App.js&utm_medium=sandpack

The solution is to wrap the `value` in a call to `useMemo`.

I've implemented that solution here where applicable, including memoizing the dependencies of those values as well.

I also cleaned up the values returned by the contexts to make them more useful and remove redundant recalculation of derived values.

### Changes

* Memoize context values
* Remove redundant getter functions
* Add helper values that can be used instead of getter functions

### DIFF
--- a/components/Panel/DelegationPanel.tsx
+++ b/components/Panel/DelegationPanel.tsx
@@ -6,6 +6,7 @@ import {
 } from "components";
 import { mobileAndUnder } from "constant";
 import { getAddress, isAddress, truncateEthAddress } from "helpers";
+import { config } from "helpers/config";
 import {
   useDelegationContext,
   useErrorContext,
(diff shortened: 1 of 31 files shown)


components/Panel/DelegationPanel.tsx
```diff
@@ -155,31 +146,35 @@ export function DelegationPanel() {
...
-                          target="_blank"
-                        >
-                          View Transaction
-                        </Link>
-                      </PendingRequestText>
-                    </PendingRequestDetailsWrapper>
-                  </AddressWrapper>
-                </PendingRequestWrapper>
-              ))}
+              pendingSentRequestsToBeDelegate.map(
```
ryanwolhuter, line 149: the only change here is formatting. kinda annoying that the diff algo sees this as a total change, but oh well

daywiss (approved):
nice update, learned something new

PR #259, opened 2023-05-08
improve: simplify loading states

ryanwolhuter:
We want it to be much simpler to figure out whether things are loading. Right now we have a lot of UI blocking going on when it doesn't need to be. There are also awkward situations where we have provided `initialData` in queries when we should not have -- this causes the `isLoading` state to never be true, which is not what we want. 

We also should move away from using the `isFetching` values from the query hooks because they cause us to show the loaders in situations where it would be better to just show the stale data instead while we refetch.


components/VoteList/VoteListItem.tsx
```diff
@@ -424,9 +426,7 @@ export function VoteListItem({
         <VoteStatusCell as={isTabletAndUnder ? "div" : "td"}>
           <VoteLabel>Vote status</VoteLabel>
           <VoteStatus>
-            {isFetching ? (
-              <LoadingSkeleton width="8vw" />
-            ) : (
+            <Loader dataToWatch={[isCommit, isReveal]} width="8vw">
```
ryanwolhuter, line 429: here's an example of watching multiple items

components/VoteList/VoteListItem.tsx
```diff
@@ -452,7 +452,7 @@ export function VoteListItem({
 }
 
 function VoteText({ voteText }: { voteText: string | undefined }) {
-  if (!voteText) return <LoadingSkeleton width="8vw" />;
+  if (voteText === undefined) return <LoadingSkeleton width="8vw" />;
```
ryanwolhuter, line 455: I realized that this would have shown a permanent loader when passed an empty string because of the falsy check

contexts/DelegationContext.tsx
```diff
@@ -40,8 +40,22 @@ export interface DelegationContextState {
   ignoreReceivedRequestToBeDelegate: (delegatorAddress: string) => void;
   terminateRelationshipWithDelegate: () => void;
   terminateRelationshipWithDelegator: () => void;
-  getDelegationDataLoading: () => boolean;
-  getDelegationDataFetching: () => boolean;
+  isLoading: boolean;
```
ryanwolhuter, line 43: the loading situation for the delegation context is a bit complicated because it involves a lot of interconnected values. Since I can't see any situations where we could benefit from treating the various ...isLoading items separately, I've added convenient values that aggregate them: isLoading, isMutating, isBusy (either loading or mutating).

I'm still passing along all of these loading states so that they could be used if need be, but I doubt we'll want to.

contexts/UserContext.tsx
```diff
@@ -35,7 +36,8 @@ export interface UserContextState {
 export const defaultUserContextState: UserContextState = {
   connectedWallet: undefined,
   account: undefined,
-  address: "",
+  address: undefined,
```
ryanwolhuter, line 39: I've opted to avoid using the empty string for the address as it leads to confusion about when the address really exists or not. We now always have an `undefined` address when it does not exist.

contexts/VotesContext.tsx
```diff
@@ -41,69 +35,73 @@ import {
...
-  encryptedVotes: {},
-  decryptedVotes: {},
-  contentfulData: {},
-  getUserDependentIsLoading: () => false,
-  getUserIndependentIsLoading: () => false,
-  getIsLoading: () => false,
-  getUserDependentIsFetching: () => false,
-  getUserIndependentIsFetching: () => false,
-  getIsFetching: () => false,
+  hasPreviouslyCommittedAll: undefined,
```
ryanwolhuter, line 76: everything starts off as undefined now

contexts/VotesContext.tsx
```diff
@@ -250,29 +160,29 @@ export function VotesProvider({ children }: { children: ReactNode }) {
     // for a delegate who committed, they will never be the voter, but will be the caller and have a vote by the delegator
     return (
       // this table finds all votes as the voter for the current account
-      (!!committedVotes[uniqueKey] &&
+      (!!committedVotes?.[uniqueKey] &&
```
ryanwolhuter, line 163: adding null checks

helpers/web3/wallet.ts
```diff
@@ -19,7 +19,7 @@ export function handleDisconnectWallet(
 export function getAccountDetails(connectedWallets?: WalletState[]) {
   const connectedWallet = connectedWallets?.[0];
   const account = connectedWallet?.accounts[0];
-  const address = account?.address ? getAddress(account.address) : "";
+  const address = account?.address ? getAddress(account.address) : undefined;
```
ryanwolhuter, line 22: undefined instead of ""

components/ErrorBanner/OldDesignatedVotingAccountWarningBanner.tsx
```diff
@@ -13,13 +13,9 @@ import {
...
 
 export function OldDesignatedVotingAccountWarningBanner() {
-  const {
-    oldDesignatedVotingAccount: {
-      isOldDesignatedVotingAccount,
-      message,
-      designatedVotingContract,
-    },
-  } = useStakingContext();
+  const { oldDesignatedVotingAccount } = useStakingContext();
```
ryanwolhuter, line 16: most of these changes are the same. We're no longer able to destructure these values directly because we can't destructure undefined. So instead we just get the value and destructure either it or an empty object. This has the effect of making the destructured values undefined if the value from the hook is undefined.

components/HowItWorks/HowItWorks.tsx
```diff
@@ -124,22 +111,20 @@ export function HowItWorks() {
...
               You have voted in{" "}
               <Strong>
-                {countCorrectVotes === undefined ||
-                countWrongVotes === undefined ? (
-                  <LoadingSkeleton width={50} />
-                ) : (
-                  formatNumberForDisplay(getTotalVotes(), { decimals: 0 })
-                )}
+                <Loader
+                  isLoading={isAnyUndefined(countCorrectVotes, countWrongVotes)}
```
ryanwolhuter, line 115: this usage is only temporary — once this is merged I can go ahead with refactoring these things into hooks, and those hooks will be able to provide these `isLoading` values in a much cleaner way.

daywiss:
just wanted to note this, in the original app when not connected:
image
in this pr:
image

contexts/DelegationContext.tsx
daywiss: makes sense, though hopefully we will be movign away from these generalized data hooks in the future

components/Loader.tsx
```diff
@@ -0,0 +1,23 @@
...
+  children,
+  error,
+  ...delegated
+}: Props) {
+  if (isError && !!error) return <>{error}</>;
+
+  if (isLoading) return <LoadingSkeleton {...delegated} />;
+
+  return <>{children}</>;
+}
```
daywiss, line 23: in your first iteration you had a way to set a default state, which was the state that showed when no wallet was connected, now it just shows loading if wallet is not connected. i know we discussed this but dont remember if this changed behavior was intentional

daywiss (approved):
gonna approve this to un block you, since i think it looks good, but behavior slightly differs from original. more of a design question, but i doubt its a problem

PR #260, opened 2023-05-09
improve: upgrade react query and add linting

ryanwolhuter:
react query has a nice eslint plugin that I was using on another branch for an upcoming PR. However, since react query v5 is going to [deprecate using the array-based arguments to `useQuery`](https://tanstack.com/query/v4/docs/react/eslint/prefer-query-object-syntax) the eslint plugin shows warnings everywhere we've used the old syntax. I've applied the automatic fix here.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -9,6 +9,7 @@
     "plugin:@typescript-eslint/eslint-recommended",
     "plugin:@typescript-eslint/recommended",
     "plugin:@typescript-eslint/recommended-requiring-type-checking",
+    "plugin:@tanstack/eslint-plugin-query/recommended",
     "prettier"
   ],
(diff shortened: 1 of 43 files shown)


.eslintrc.json
```diff
@@ -22,6 +23,8 @@
         "varsIgnorePattern": "^_",
         "caughtErrorsIgnorePattern": "^_"
       }
-    ]
+    ],
+    "@tanstack/query/exhaustive-deps": "warn",
```
ryanwolhuter, line 27: we can't set this one to `error` yet because there are some places where our contract instances are dependencies of our functions and would therefore be expected to be placed in the dependency array. This won't work because the instances are not json serializable. This is one of the problems that viem aims to fix, so I'm looking forward to moving over to that once its ready.

PR #264, opened 2023-05-16
improve: refactor voterdapp hooks

daywiss:
## motivation
this is a continuation of #256 .  separated pr so that diffs can be compared in case conflicts were not resolved correctly

## changes
This basically rebases 256 on top of the latest master, merging hook refactor and loading. no other changes other than changes to fix issues with basic qa


daywiss:
refactor seems to be having some issues. shoudl take a look at that first

PR #267, opened 2023-05-18
fix: use object as param to prevent accidentally using wrong values

ryanwolhuter:
Some of our mutations had an optional param for `errorOrigin`. We then updated those mutations to also take the `address` to use as the first param. But because the second param was optional, and the address is just a string, typescript did not flag any errors in the places where we called the mutations with the string of the `errorOrigin` instead of the string of the `address`. So essentially we were trying to use the string "claim" or the string "stake" as the address.

I've updated these mutations to take object arguments instead, so that this ambiguity cannot happen in future.

### DIFF
--- a/components/Panel/ClaimPanel.tsx
+++ b/components/Panel/ClaimPanel.tsx
@@ -27,8 +27,14 @@ export function ClaimPanel() {
   const { isDelegate, delegatorAddress, outstandingRewards } =
     useDelegationContext();
   const { address } = useAccountDetails();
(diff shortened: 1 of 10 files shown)


daywiss (approved):
ive tested
stake, unstake, claim and stake, claim
delegated claim and stake - which does have slight problems with correctly resetting claimable amount, but i think thats unrelated to this pr. the call works.

we can wait to merge this until after vote

PR #268, opened 2023-06-21
feat: implement new table view style

ryanwolhuter:
Update the votes table to look like the table in the oracle dapp:

 - chevron icon instead of more details
 - title can wrap lines
 - highlight title red on hover

Doing this required separating the mobile and desktop views into their own components. The state logic remains unchanged.

I installed and configured tailwindcss to achieve this, which enabled the deletion of a lot of style code.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -7,6 +7,11 @@
   resolved "https://registry.yarnpkg.com/@adobe/css-tools/-/css-tools-4.2.0.tgz#e1a84fca468f4b337816fcb7f0964beb620ba855"
   integrity sha512-E09FiIft46CmH5Qnjb0wsW54/YQd69LsxeKUOWawmws1XWvyFGURnAChH0mlr7YPFR1ofwvUQfcL0J3lMxXqPA==
 
+"@alloc/quick-lru@^5.2.0":
+  version "5.2.0"
+  resolved "https://registry.yarnpkg.com/@alloc/quick-lru/-/quick-lru-5.2.0.tgz#7bf68b20c0a350f936915fcae06f58e32007ce30"
+  integrity sha512-UrcABB+4bUrFABwbluTIBErXwvbsU/V7TZWfmbgJfbkwiBuziS9gxdODUyuiecfdGQ85jglMW6juS3+z5TsKLw==
+
 "@ampproject/remapping@^2.1.0", "@ampproject/remapping@^2.2.0":
   version "2.2.0"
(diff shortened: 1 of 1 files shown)


components/VoteList/VoteList.tsx
```diff
@@ -1,36 +1,41 @@
...
 }
-export function VoteList({ headings, rows }: Props) {
+export function VoteList({ data, activityStatus }: Props) {
   const { width } = useWindowSize();
 
   if (!width) return null;
 
   const isTabletAndUnder = width <= tabletMax;
 
+  if (isTabletAndUnder)
```
ryanwolhuter, line 20: this is the most important change. we now have two different components for mobile and desktop. this allowed me to separate the styling for the table stuff from the mobile list stuff

components/VoteList/useVoteListItem.tsx
```diff
@@ -0,0 +1,326 @@
...
+import OSnap from "public/assets/icons/osnap.svg";
+import Polymarket from "public/assets/icons/polymarket.svg";
+import UMAGovernance from "public/assets/icons/uma-governance.svg";
+import UMA from "public/assets/icons/uma.svg";
+import { CSSProperties, useEffect, useState } from "react";
+import removeMarkdown from "remove-markdown";
+import { DropdownItemT } from "types";
+import { VoteListItemProps } from "./shared.types";
+
+export function useVoteListItem({
```
ryanwolhuter, line 26: all of the state that lived in the `VoteListItem` has been copied here and left unchanged. This is now also used by the `VoteTableRow` component and the behavior is unchanged. the only thing that is removed is some logic for calculating table cell widths which are now redundant because we use an automatic table layout

components/Votes/ActiveVotes.tsx
```diff
@@ -311,33 +309,29 @@ export function ActiveVotes() {
     selectVote(undefined, vote);
   }
 
+  const data = entriesToShow?.map((vote, index) => ({
```
ryanwolhuter, line 312: now instead of passing in data rows as react elements, we just pass in the data and do the iteration in the `VoteList` component instead

components/VoteList/VoteListItem.tsx
```diff
@@ -6,386 +6,72 @@ import {
...
+    getYourVote,
+    showCorrectVote,
+    getCorrectVote,
+    showVoteStatus,
+    isLoading,
+    getRelevantTransactionLink,
+    isDirty,
+    origin,
+    moreDetailsAction,
+  } = useVoteListItem(props);
```
ryanwolhuter, line 46: state logic is the same. just moved to this hook

daywiss (approved):
did some testing, looks good

PR #269, opened 2023-07-03
fix: update onboard deps, walletconnect v2

daywiss:
## motivation
wallet connect stopped working when v1 was deprecated. 

## changes
this bumps all onboard versions, adds wallet connect project id, and new configuration. needs testing for both wallet connect and non wallet connect instances


ryanwolhuter (approved):
Just make sure to add the envs to both deployments

PR #270, opened 2023-07-11
fix: disable staking max button with no wallet connected

daywiss:
## motivation
when wallet not connected, there was a way to enable the stake button by messing with the input box. this leads to errors on the app.

## changes
this makes sure we know to disable the staking inputbox and other buttons when no wallet is connected. 

image


components/Panel/StakeUnstakePanel/Stake.tsx
```diff
@@ -62,6 +64,10 @@ export function Stake({
     );
   }
 
+  function isInputDisabled() {
```
ryanwolhuter, line 67: why not just use the value directly instead of using a function?

daywiss: no reason really, just noticed there were other similar functions so following the pattern, also its better for clarity

PR #272, opened 2023-08-20
fix: discord evidence rationale link

md0x:
Changes proposed in this PR:
- Fix the link to the discord threads by using the `message.thread.name` instead of `message.content`. 
  - Explanation: In some cases when editing the thread title after creation, `message.content` returned by discord API still returns the old title but `message.thread.name` is up to date. That's why we always need to use `message.thread.name`


daywiss (approved):
thanks for looking into this, good find

PR #273, opened 2023-08-22
improve: change ownership pronoun

james-a-morris:
When talking about something that belongs to a single person or thing, I believe we'd want to use "its" to show possession. 

I.e. Your delegator is staking X of their Y UMA tokens.

In the above example, _their_ refers to _Your delegator_, and since there's only one delegator, it should be _its_.

### DIFF
--- a/components/HowItWorks/HowItWorks.tsx
+++ b/components/HowItWorks/HowItWorks.tsx
@@ -108,7 +108,7 @@ export function HowItWorks() {
                   {formatNumberForDisplay(stakedBalance)}
                 </Loader>
               </Strong>{" "}
-              {isDelegate ? "of their" : "of your"}{" "}
+              {isDelegate ? "of its" : "of your"}{" "}
               <Strong>
                 <Loader
                   isLoading={


daywiss:
@james-a-morris thanks for making this a pr.  does it seem weird though that when you say "its" tokens its like referring to something that is not a person, something inanimate maybe?  "their" gives the impression its a person.  that said im not opposed to this change, what do you think?

`Your delegator is staking X of their Y UMA tokens.`
vs 
`Your delegator is staking X of its Y UMA tokens.`

mrice32:
> @james-a-morris thanks for making this a pr. does it seem weird though that when you say "its" tokens its like referring to something that is not a person, something inanimate maybe? "their" gives the impression its a person. that said im not opposed to this change, what do you think?
> 
> `Your delegator is staking X of their Y UMA tokens.` vs `Your delegator is staking X of its Y UMA tokens.`

I really appreciate you keeping an eye on the dapp copy @james-a-morris. +1!

Not a big issue either way, but I agree with @daywiss here. I think `its` would imply that the delegator is a thing, not a person. However, the delegator is commonly assumed to be a person. Despite they/their being a plural pronoun, it's long been acceptable to use it as a pronoun for an unknown or unidentified singular 3rd person, especially outside of formal writing.

Example 1:

"My friend sends their regards." is correct.
"My friend sends its regards." is incorrect because its implies that friend is a thing, not a person.

Example 2:

"The door had its handle replaced." is correct.
"The door had their handle replaced." is incorrect because it implies that the door is a person, rather than a thing.

james-a-morris:
Sounds good to me! Thanks everyone for clarifying that the delegator is a human operator rather than an entity 🙂

PR #274, opened 2023-08-23
fix: option parsing for polymarket requests

daywiss:
## motivation
since polymarket changed ancillary data format, we need to update dynamic option parsing

## changes
 this updates the regex to account for a period where a comma used to be


mrice32 (approved):

daywiss:
this was fixed in diff pr

PR #276, opened 2023-08-23
fix: make user re-select their vote when a vote rolls

ryanwolhuter:
There is a bug where the app does not recognize that a rolled vote is a new vote. it repopulates the form field with the user's encrypted vote from the previous round. 

We want the vote to be treated as a new vote. In my opinion it does not make sense to automatically assume that the user wants to vote the same as the previous time for a rolled vote anyway, since by the nature of a rolled vote it is likely that the user has changed their mind. At the very least they should consider their new vote before committing.

To fix this behavior we should filter encrypted votes by round id.

### DIFF
--- a/contexts/VotesContext.tsx
+++ b/contexts/VotesContext.tsx
@@ -18,6 +18,7 @@ import {
   useRevealedVotes,
   useUpcomingVotes,
   useUserVotingAndStakingDetails,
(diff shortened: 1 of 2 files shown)


daywiss (approved):
thanks for looking into this, not sure why this was never noticed before

PR #277, opened 2023-09-01
improve: commify uma large displays of tokens in vote results

daywiss:
## motivation
sometimes it's interesting to follow the voter numbers to gauge approx votes revealed vs. remaining, but my brain short circuits when I try to parse long numbers without a separator.
ethers has a helper called [commify()](https://docs.ethers.org/v5/single-page/#/v5/api/utils/display-logic/-%23-utils-commify) to do this automagically - could that be an easy way to format it?What's the reason for the improvement?

## changes
commify the 2 places where we display this in vote results


ryanwolhuter (approved):
👍🏻

PR #278, opened 2023-10-19
improve: improve osnap vote rendering

daywiss:
## motivation
we arent correctly detecting osnap assertions, and not showing relevant data

## changes
1. this makes it so we correctly detect assertions ( we were looking for a string that doesnt exist on these osnap assertions )
2. we decode the claim now in details
3. we have the correct options, previously the options were missing, so votes should be woking

with more time we need to fix showing the correct title, and parsing the claim data similar to OO.

### DIFF
--- a/components/Panel/VotePanel/Details.tsx
+++ b/components/Panel/VotePanel/Details.tsx
@@ -24,7 +24,7 @@ import Time from "public/assets/icons/time-with-inner-circle.svg";
 import { useState } from "react";
 import ReactMarkdown from "react-markdown";
 import styled from "styled-components";
-import { VoteT } from "types";
+import { VoteT, SupportedChainIds, OracleTypeT } from "types";
 import { PanelSectionTitle } from "../styles";
 import { ChainIcon } from "./ChainIcon";
 import { OoTypeIcon } from "./OoTypeIcon";
@@ -43,12 +43,12 @@ export function Details({
   augmentedData,
   assertionChildChainId,
   assertionAsserter,
-  claim,
-}: VoteT & { claim: string | undefined }) {
+}: VoteT) {
   const [showDecodedAdminTransactions, setShowDecodedAdminTransactions] =
     useState(false);
   const [showRawAncillaryData, setShowRawAncillaryData] = useState(false);
   const [showRawClaimData, setShowRawClaimData] = useState(false);
+  const claim = augmentedData?.optimisticOracleV3Data?.claim;
   const isClaim = !!claim;
   const showAncillaryData = !isClaim;
   function toggleShowDecodedAdminTransactions() {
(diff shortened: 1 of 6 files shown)


components/Panel/VotePanel/Details.tsx
```diff
@@ -116,8 +118,16 @@ export function Details({
...
-          <ChainIcon chainId={augmentedData?.originatingChainId} />
-          <OoTypeIcon ooType={augmentedData?.originatingOracleType} />
+          <ChainIcon
+            chainId={
+              augmentedData?.originatingChainId as SupportedChainIds | undefined
+            }
+          />
+          <OoTypeIcon
+            ooType={
+              augmentedData?.originatingOracleType as OracleTypeT | undefined
```
daywiss, line 128: these have to get casted because we are relying on a superstruct type which doesnt have the nuances for these special types. perhaps this can be improved in another pass

components/Panel/VotePanel/Details.tsx
```diff
@@ -43,12 +43,12 @@ export function Details({
   augmentedData,
   assertionChildChainId,
   assertionAsserter,
-  claim,
```
daywiss, line 46: turns out claim is already provided through a serverless call to get augmented data, so we prefer that, vs this tacked on data which doesnt seem to be workign

components/Panel/VotePanel/Details.tsx
```diff
@@ -101,7 +101,9 @@ export function Details({
     augmentedData.originatingOracleType
       ? makeTransactionHashLink(
           `${
-            supportedChains[augmentedData.originatingChainId]
+            supportedChains[
+              augmentedData.originatingChainId as SupportedChainIds
```
daywiss, line 105: see casting justification in next code block

helpers/voting/getVoteMetaData.ts
```diff
@@ -21,9 +21,10 @@ export function getVoteMetaData(
   decodedAncillaryData: string,
   umipDataFromContentful: ContentfulDataT | undefined
 ): VoteMetaDataT {
-  const isAssertion = ["assertionId:", "ooAsserter:", "childChainId:"].every(
-    (lookup) => decodedAncillaryData.includes(lookup)
```
daywiss, line 25: this "childChainId" check was causing us to miss that this was an assertion. it doesnt seem to be included in the ancillary data for this vote. figured this is still strict enough to not have false positives

pages/api/augment-request.ts
```diff
@@ -79,7 +79,7 @@ function constructOoUiLink(
   const subDomain = Number(chainId) === 5 ? "testnet." : "";
   return `https://${subDomain}oracle.uma.xyz/request?transactionHash=${txHash}&chainId=${chainId}&oracleType=${castOracleNameForOOUi(
     oracleType
-  )}`;
+  )}&eventIndex=`;
```
daywiss, line 82: for some reason if this isnt included, we dont link correctly to the OO UI. even if its left blank like it is here we can deduce the correct request in the UI

Reinis-FRP (approved):

PR #280, opened 2023-12-11
fix: votes slashed pagination

md0x:
Changes proposed in this PR:
- Handles pagination in votes slashed in users query to fetch vote history participation


daywiss (approved):
Thanks for looking into this, nice find

PR #282, opened 2023-12-20
feat: ensure we fetch all past votes

gsteenkamp89:
## motivation

thegraph server only returns 1000 documents at a time and does not support standard PageInfo features like cursors for pagination. So we might get into a situation where we think we are fetching all documents but we're only fetching this first 1000. 

It seems unlikely that we'll ever have more than 1000 upcoming votes so I'm only implementing this logic for `pastVotes`

In this solution all we're doing is fetching 1000 at a time until we get less that 1000 then we know we've got them all. 

closes #UMA-2125

### DIFF
--- a/helpers/util/fetchAllDocuments.ts
+++ b/helpers/util/fetchAllDocuments.ts
@@ -1,0 +1,40 @@
+import request from "graphql-request";
+
+/**
+ * Thegraph can only return a maximum of 1000 documents at a time and does not support pagination features whereby the client can determine the total amount of documents in the store.
+ * This helper requests 1000 documents at a time until there are no more ensuring we don't miss any
+ * @param endpoint the endpoint for your graphql request
+ * @param query your graphql request document
+ * @param dataKey the key for the data you want. this is the root key of your graphql document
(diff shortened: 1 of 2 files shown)


helpers/util/fetchAllDocuments.ts
```diff
@@ -0,0 +1,40 @@
...
+  endpoint: string,
+  query: string,
+  dataKey: keyof RequestType,
+  pageSize = 1000
+) {
+  let allData = new Array<RequestType[keyof RequestType][number]>();
+  let skip = 0;
+  let continueFetching = true;
+
+  while (continueFetching) {
```
daywiss, line 26: could probably make this a bit more elegant with a do while loop which could get rid of a var
```
do {
  //... code
  skip += pageSize;
} while(response[dataKey].length < pageSize)
```

daywiss (approved):

helpers/util/fetchAllDocuments.ts
gsteenkamp89: Good idea!

gsteenkamp89: okay yeah the do while loop does seem cleaner 👍

PR #285, opened 2023-12-27
feat: add support for neg risk market

daywiss:
## motivation
we need voter dapp to support non standard yes no market which only has yes, no and too early ( missing p3)

## changes
this adds a lookup for this new market, but also orders parsing to go from most specific to least specific, since each option type is a subset of another


helpers/voting/polymarket.ts
```diff
@@ -125,6 +136,28 @@ export function maybeMakePolymarketOptions(
...
+      {
+        label: "Custom",
+        value: "custom",
+      },
+    ];
+  }
+
+  if (
+    decodedAncillaryData.includes(options3.resData) &&
+    decodedAncillaryData.includes(options3.corresponds)
```
daywiss, line 148: this is the new type which only has yes no and early request

gsteenkamp89 (approved):

PR #286, opened 2024-01-02
improve: remove unused chain, disable creating providers if no envs supplied

daywiss:
## motivation
theres a code smell where we are initializing providers with potentially undefined data

## changes
this removes boba which we no longer support and filters out provider urls that are undefined before creating instances


gsteenkamp89 (approved):
👍

PR #287, opened 2024-01-19
feat: osnap data augmentation

md0x:
Proposed changes in this Pull Request:
- Temporary fix to properly render the oSnap votes.
Note: This should be further enhanced by moving the oSnap data fetching to a higher level in the component hierarchy. This will help in avoiding code duplication, improving data fetching efficiency, etc.


daywiss (approved):
thank you!

PR #289, opened 2024-01-29
clamp lines to 2 in vote list titles

gsteenkamp89:
## motivation

We want to use CSS to truncate text if it spans more than 2 lines. We think 2 lines is more than enough space for a title, but in the rare case that it is larger we don't want to break the layout.

## screenshots

Screenshot 2024-01-29 at 18 29 55

closes UMA-2311


daywiss (approved):
looks good

PR #290, opened 2024-01-30
Gerhard/uma 2333 voting dapp discord discussions not wokring

gsteenkamp89:
## motivation

Discord threads not showing up in side panel.

Minor bug introduced a while back causing a reference error at runtime.
Using Discord's official API types to further shield us from attempting to access nullible fields

### DIFF
--- a/pages/api/discord-thread.ts
+++ b/pages/api/discord-thread.ts
@@ -99,7 +99,7 @@ async function fetchDiscordThread(
 ): Promise<VoteDiscussionT> {
   // First, fetch all messages in the evidence rational channel.
   const threadMsg = await getDiscordMessages(evidenceRationalDiscordChannelId);
-  console.log(threadMsg);
+
   // Then, extract the timestamp from each message and for each timestamp relate
   // it to the associated threadId.
   const timeToThread: { [key: string]: string | undefined } = {};
--- a/pages/api/discord-thread.ts
+++ b/pages/api/discord-thread.ts
@@ -9,6 +9,7 @@ import {
 } from "types";
 
 import * as ss from "superstruct";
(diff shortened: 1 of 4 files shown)


pages/api/discord-thread.ts
```diff
@@ -51,16 +52,20 @@ function replaceMentions(
...
+
+function replaceEmbeds(message: string, embeds: APIEmbed[]) {
   let result = message;
   for (const embed of embeds) {
-    result = result.replace(`(${embed.url})`, `[${embed.title}](${embed.url})`);
+    if (embed?.url && embed?.title) {
+      result = result.replace(
+        `(${embed.url})`,
+        `[${embed.title}](${embed.url})`
+      );
```
gsteenkamp89, line 63: if null we just return the string as is

pages/api/discord-thread.ts
```diff
@@ -94,22 +99,25 @@ async function fetchDiscordThread(
...
   const threadMsg = await getDiscordMessages(evidenceRationalDiscordChannelId);
+
   // Then, extract the timestamp from each message and for each timestamp relate
   // it to the associated threadId.
-  const timeToThread: { [key: string]: string } = {};
+  const timeToThread: { [key: string]: string | undefined } = {};
   threadMsg.forEach((message) => {
-    const time = extractValidateTimestamp(message.thread.name);
-    if (time) timeToThread[time.toString()] = message.thread.id;
+    const time = extractValidateTimestamp(message.content);
```
gsteenkamp89, line 107: this was causing the crash, when `name` was undefined

daywiss: maybe we need to do message.thread.name ?? message.content

daywiss (approved):

PR #291, opened 2024-02-02
improve: change all headers in discord comments to bold

daywiss:
## motivation
discord comments are rendered as markdown, its possible to make really large text in comments by using header tag #.  we want to remove this ability

before
image

after
image

## changes
this uses a regex to identify these headers and turn them bold.  also fix some bold formatting as well


gsteenkamp89 (approved):
nice

PR #296, opened 2024-03-07
feat: support sepolia testnet

Reinis-FRP:
Adds support for Sepolia testnet

Fixes: https://linear.app/uma/issue/UMA-2448/update-voter-dapp-to-support-sepolia


daywiss (approved):
looks good

PR #297, opened 2024-03-07
fix: use og graph to determine if is og vote

daywiss:
## motivation
Non osnap assertion votes are causing app to crash

## changes
This uses the subgraph to determine if something is a og vote, rather than the ancillary data


Reinis-FRP (approved):
nice!

PR #305, opened 2024-04-24
fix: differentiate polymarket from other yes no queries

daywiss:
## Motivation
We have another yes no test request which is not from polymarket but is incorrectly being labeled as such

## changes
this adds known polymarket addresses and child chain to label polymarket requests

image


gsteenkamp89 (approved):
🤷‍♂️ hope it works

PR #315, opened 2024-07-02
reset outstanding rewards if no delegate wallet connected

gsteenkamp89:
## To test

1. Connect delegate wallet, should have some outstanding rewards
2. Disconnect wallet => outstanding rewards should read `0`
3. Connect delegate wallet again (outstanding rewards should appear again) then switch to non-delegate wallet, outstanding rewards should now read `0`


daywiss (approved):
looks good

PR #316, opened 2024-07-11
feat: improve wording around commited/revealed votes

gsteenkamp89:
## motivation

Some users have noticed it is a bit unclear how many users have committed or revealed since the UI previously only shows amount of **votes**. 
Now we show how many tokens committed and revealed. we also show the percentage of users that revealed.

## screenshots

Screenshot 2024-07-11 at 16 23 51


daywiss:
noticed that past votes component is not loading for some reason

PR #317, opened 2024-07-23
fix: use the graph for augmented data

daywiss:
# Motivation
augment request which is responsible for finding metadata about transactions has been failing

# changes
This updates augment request to use the graph and to only query for a single request as its displayed, this speeds up calls, improves caching, and hopefully stops erroring and timing out

oracle v3:
image
oracle v2:
image
skinny oo:
image

### DIFF
--- a/components/Panel/VotePanel/Details.tsx
+++ b/components/Panel/VotePanel/Details.tsx
@@ -14,7 +14,7 @@ import {
 } from "helpers";
 import { config } from "helpers/config";
 import { useOptimisticGovernorData } from "hooks/queries/votes/useOptimisticGovernorData";
-import { useAssertionClaim } from "hooks";
+import { useAssertionClaim, useAugmentedVoteData } from "hooks";
 
 import AncillaryData from "public/assets/icons/ancillary-data.svg";
 import Chat from "public/assets/icons/chat.svg";
@@ -43,7 +43,6 @@ export function Details({
   discordLink,
   decodedAdminTransactions,
   umipOrUppLink,
-  augmentedData,
   assertionChildChainId,
   assertionAsserter,
   assertionId,
@@ -68,6 +67,13 @@ export function Details({
   function toggleShowRawClaimData() {
     setShowRawClaimData(!showRawClaimData);
   }
+
+  const augmentedDataResponse = useAugmentedVoteData({
+    time,
+    identifier: decodedIdentifier,
+    ancillaryData: ancillaryData,
+  });
+  const augmentedData = augmentedDataResponse.data;
 
   function makeOoRequestLink() {
     if (!augmentedData?.ooRequestUrl) return;
--- a/hooks/queries/votes/useAugmentedVoteData.ts
+++ b/hooks/queries/votes/useAugmentedVoteData.ts
@@ -1,27 +1,32 @@
+import assert from "assert";
 import { useQuery } from "@tanstack/react-query";
 import { augmentedVoteDataKey } from "constant";
 import { getAugmentedVoteData } from "web3";
-import { useActiveVotes } from "./useActiveVotes";
-import { usePastVotes } from "./usePastVotes";
-import { useUpcomingVotes } from "./useUpcomingVotes";
 
-export function useAugmentedVoteData() {
-  const { data: activeVotes } = useActiveVotes();
-  const { data: upcomingVotes } = useUpcomingVotes();
-  const { data: pastVotes } = usePastVotes();
-
-  const allVotes = { ...activeVotes, ...upcomingVotes, ...pastVotes };
-
-  const queryResult = useQuery({
+export function useAugmentedVoteData(
+  params: Partial<{
+    ancillaryData: string;
+    time: number;
+    identifier: string;
+  }>
+) {
+  return useQuery({
     queryKey: [
       augmentedVoteDataKey,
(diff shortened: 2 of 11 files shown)


components/Panel/VotePanel/Details.tsx
```diff
@@ -69,6 +68,13 @@ export function Details({
     setShowRawClaimData(!showRawClaimData);
   }
 
+  const augmentedDataResponse = useAugmentedVoteData({
+    time,
+    identifier: decodedIdentifier,
+    ancillaryData: ancillaryData,
+  });
+  const augmentedData = augmentedDataResponse.data;
+
```
daywiss, line 77: this used to be fetched in the vote context for all votes. now we move this to the render component, and only fetch a single augmented data per vote when we are viewing the details pane.  much more efficient, and easier to cache

hooks/queries/votes/useAugmentedVoteData.ts
```diff
@@ -1,27 +1,32 @@
...
-
-  const allVotes = { ...activeVotes, ...upcomingVotes, ...pastVotes };
-
-  const queryResult = useQuery({
+export function useAugmentedVoteData(
+  params: Partial<{
+    ancillaryData: string;
+    time: number;
+    identifier: string;
+  }>
```
daywiss, line 11: previously when calling this, it would fetch all augmented data for all votes.  now we pass params to specify the single vote we care about

pages/api/augment-request-gql.ts
```diff
@@ -0,0 +1,408 @@
...
+    );
+
+  if (!successResult) {
+    throw new Error("All requests failed");
+  }
+
+  return successResult?.value;
+}
+
+async function augmentRequest(request: RequestBody): Promise<ResponseBody> {
```
daywiss, line 379: entry point function is a  lot cleaner now. this works by getting the l1 vote data, and figuring out where the chain id of the originating request came from using the ancillary data ( includes child chain id).  Then we query every known type of OO on that chain and hopefully find one that returns valid data.  This lets us know what type of oracle it was requested from.

pages/api/augment-request-gql.ts
```diff
@@ -0,0 +1,408 @@
...
+}
+async function oov3Query({
+  time,
+  identifier,
+  chainId,
+}: OracleQueryParams): Promise<ResponseBody> {
+  const subgraph = getSubgraphConfig("Optimistic Oracle V3", chainId);
+  const query = gql`
+    query oov3Query($assertionTimestamp: Int!) {
+      assertions(where: { assertionTimestamp: $assertionTimestamp }) {
```
md0x, line 170: `assertionTimestamp` doesn't uniquely identify an assertion
We should use the `ancillaryData` from the VotingV2 price request, parse it to extract the `assertionId` and query the subgraph by `assertionId`

See how the ancillaryData is calculated in the OOv3 [here](https://github.com/UMAprotocol/protocol/blob/454cd8aed43094c94916d6225e4c52403d8bdf19/packages/core/contracts/optimistic-oracle-v3/implementation/OptimisticOracleV3.sol#L394)

pages/api/augment-request-gql.ts
```diff
@@ -0,0 +1,408 @@
...
+  }
+}
+async function ooSkinnyQuery({
+  time,
+  identifier,
+  chainId,
+}: OracleQueryParams): Promise<ResponseBody> {
+  const subgraph = getSubgraphConfig("Skinny Optimistic Oracle", chainId);
+  const query = gql`
+    query skinnyQuery($proposalTimestamp: Int!) {
```
md0x, line 116: 2 price requests can have identical`proposalTimestamp` s

We need to filter in the subgraph by: ancillaryData, time, identifier

pages/api/augment-request-gql.ts
daywiss: ive updated this to be more specific

pages/api/augment-request-gql.ts
daywiss: updated this

package.json
```diff
@@ -24,6 +24,7 @@
     "@tanstack/react-query": "^4.29.5",
     "@tanstack/react-query-devtools": "^4.29.6",
     "@uma/contracts-frontend": "^0.4.20",
+    "@vercel/kv": "^2.0.0",
```
gsteenkamp89, line 27: this can be removed

pages/api/augment-request-gql.ts
```diff
@@ -0,0 +1,465 @@
...
+import { isSupportedChainId, getSubgraphConfig } from "./_common";
+import { ethers } from "ethers";
+
+function encodeHexString(str: string): string {
+  return ethers.utils.hexlify(ethers.utils.toUtf8Bytes(str));
+}
+const debug = !!process.env.DEBUG;
+const VoteSubgraphURL =
+  process.env.NEXT_PUBLIC_GRAPH_ENDPOINT ??
+  "https://api.thegraph.com/subgraphs/name/md0x/mainnet-voting-v2";
```
gsteenkamp89, line 15: is it possible to pull this from our config (constants) file?

gsteenkamp89 (approved):
lgtm 👍

md0x (approved):
LGTM! This should save a lot of rpc calls

PR #321, opened 2024-07-31
add link to polymarket

gsteenkamp89:
## motivation
it would be useful to link to the relevant prediction market from the voter dapp.

Screenshot 2024-07-31 at 13 00 20


hooks/queries/votes/usePolymarketData.ts
```diff
@@ -0,0 +1,57 @@
...
+  const res = await fetch(url);
+  const search = (await res.json()) as PolymarketResponse;
+  if (!search.length) {
+    throw new Error("Unable to find market.");
+  }
+  const data = search[0];
+
+  return {
+    ...data,
+    link: `${POLYMARKET_BASE_URL}/event/${data.slug}`,
```
daywiss, line 44: you might just want to verify that data.slug exists, or just throw an error if it doesnt, since you are casting the response

daywiss (approved):
small nit otherwise looks good

PR #325, opened 2024-08-05
use new committed tokens counter from graph

gsteenkamp89:
## motivation

Since updates to the subgraph [here](https://github.com/UMAprotocol/subgraphs/pull/80), we are tracking the tokens staked by each committer, when they commit their vote. We now use this value to calculate the percentage of tokens that revealed.

Screenshot 2024-08-05 at 11 53 25


daywiss (approved):
nice improvement

PR #328, opened 2024-08-30
fix: increase subgraph query limits for commits and reveals

daywiss:
# motivation
We are only ever seeing max 100 commit addresses

# changes
apparently gql query limits you to 100 returned values, so we increased the limit, which should hold us over for a while


gsteenkamp89 (approved):
should do for now 👍

PR #332, opened 2024-10-05
fix: increase message query size

daywiss:
# motivation
this discussion is missing the first X number of posts `Will Israel invade Lebanon in September?`

# changes 
the discord api returns posts in reverse order, ie newest first. and theres a limit on number of messages that get returned. we can modify this to a max of 100, which fixes the issue for now, but we need to spend more time to add pagination to this call.

you can see in discussion tab in this pr DOMAH is first which matches discussion in discord evidence rationale
vs the prod app it shows someone else


gsteenkamp89 (approved):
nice one thanks for jumping on this

PR #334, opened 2024-10-06
fix: rework paging logic in discord

daywiss:
# motiviation
we just fixed pagin in discord, now discussion is loading forever due to rate limiting

# changes
- do not paginate thread lookups since there are so many old threads we can discard
- adjusted querying logic to prevent an extra query by ending early if messages returned is less than limit
- added sleep between loops
- next version bumped
- use get vs post to ensure caching working on serverless call
- refresh 20 sec vs on focus

### DIFF
--- a/pages/api/discord-thread.ts
+++ b/pages/api/discord-thread.ts
@@ -52,25 +52,35 @@ export async function discordRequest(endpoint: string) {
   return (await res.json()) as RawDiscordThreadT;
 }
 
+export async function getDiscordMessages(threadId: string, limit = 100) {
+  return discordRequest(`channels/${threadId}/messages?limit=${limit}`);
+}
+function sleep(ms: number): Promise<void> {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
 // fetch 100 at a time until limit is reached
-export async function getDiscordMessages(threadId: string) {
+export async function getDiscordMessagesPaginated(
+  threadId: string,
+  limit = 100
+) {
   let allMessages: RawDiscordThreadT = [];
-  const limit = 100;
   let lastMessageId: string | undefined = undefined;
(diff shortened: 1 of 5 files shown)


pages/api/discord-thread.ts
```diff
@@ -136,8 +147,11 @@ async function fetchDiscordThread(
   l1Request: L1Request
 ): Promise<VoteDiscussionT> {
   // First, fetch all messages in the evidence rational channel.
-  const threadMsg = await getDiscordMessages(evidenceRationalDiscordChannelId);
-
+  // we dont really need to go back far in history
+  const threadMsg = await getDiscordMessages(
```
daywiss, line 151: this is the main change. we dont paginate old threads, since there are so many, this reduces the number of fetches with pagination.  we dont need to go back more than 50 threads back

pages/api/discord-thread.ts
```diff
@@ -52,25 +52,35 @@ export async function discordRequest(endpoint: string) {
...
-      break; // No more messages to fetch
-    }
     allMessages = allMessages.concat(messages);
 
+    if (messages.length < limit) {
+      break; // No more messages to fetch
+    }
     lastMessageId = messages[messages.length - 1]?.id; // Get the last message ID
-  }
+    await sleep(50);
```
daywiss, line 82: small sleep statement between loops just in case

pages/api/discord-thread.ts
mrice32: Does this mean that if you go back to 50 votes back, the discord threads will no longer show up?

If there are 51 votes, will this also break?

daywiss: thats right, it will only go back so far

gsteenkamp89 (approved):

PR #341, opened 2025-01-22
fix for polymarket's new response structure

gsteenkamp89:
closes UMA-2826

## Motivation
 
PM changed the shape of their API response, so we were unable to show a link to the PM market for a particular dispute.


daywiss (approved):
great find!

PR #345, opened 2025-02-07
fix: voting history

md0x:
Changes proposed in this PR:
- Fixes the voting history issues
- Depends on the new version of the voting v2 subgraph https://github.com/UMAprotocol/subgraphs/pull/85

Due to TheGraph working memory limitations we need to show a "pending" chip for recent past votes earnings. These are updated once the voter interacts with the contract.

image


md0x:
> I see no changes to our queries to the subgraph. Is that expected?

Yes, I didn’t need to change the queries after all.

gsteenkamp89 (approved):

PR #348, opened 2025-02-11
Feat/multiple values UI

gsteenkamp89:


daywiss (approved):
looks good, i think we can fix any issues that come up in future prs:
1. cant see the entered values in reveal phase
2. results in donut chart needs to be parsed somehow
3. past votes not parsed right 

image

image
02b127240f7f)
image

PR #356, opened 2025-03-07
add endpoint for pm market link

gsteenkamp89:
closes UMA-2863
We strip all tags (childChainId, ooRequester and childRequester) from DVM ancillary data, leaving only original ancillary data + initializer. we hash this to find the questionId, replicating Polymarket's logic.

### DIFF
--- a/pages/api/get-polymarket-link.ts
+++ b/pages/api/get-polymarket-link.ts
@@ -35,13 +35,13 @@ export default async function handler(
     }>;
 
     const slug = data?.[0]?.events?.[0]?.slug;
-
-    console.log(`Found polymarket slug for question ID: ${questionId}`, slug);
-
     response.setHeader("Cache-Control", "max-age=0, s-maxage=2592000");
     response.status(200).send({ slug });
   } catch (e) {
-    console.error(e);
+    console.error("Error fetching Polymarket link", {
+      at: "api/get-polymarket-link",
+      cause: e,
+    });
 
     response.status(e instanceof HttpError ? e.status : 500).send({
       message: "Server error",


helpers/validators.ts
```diff
@@ -0,0 +1,8 @@
+import { isHexString } from "ethers/lib/utils";
+import { define } from "superstruct";
+
+export function hexString() {
+  return define<`0x${string}`>("hexString", (value) => {
+    return isHexString(value);
+  });
+}
```
daywiss, line 8: interesting never seen this used before. so does this enforce 0x at the start?

pages/api/get-polymarket-link.ts
```diff
@@ -0,0 +1,67 @@
...
+    const tx = await provider.getTransactionReceipt(hash);
+
+    if (!tx) {
+      throw new HttpError({
+        status: 404,
+        message: `Transaction not found: Unable to find transaction receipt for hash ${hash}`,
+      });
+    }
+
+    const log = tx.logs.find((log) => log.topics?.[0] === topicHash);
```
daywiss, line 40: safe to assume everything lowercase?

helpers/validators.ts
gsteenkamp89: It checks 0x at the start and also check if all nibbles are valid hex and if there are an even number of them I think

daywiss (commented):
pm link doesnt show up on all votes currently, seeing this error:
image

gsteenkamp89:
Keeping this in draft until we can find a better solution. 
Currently this method only produces a link in ~70% of cases

pages/api/get-polymarket-link.ts
```diff
@@ -0,0 +1,51 @@
...
+      })}`
+    ).then((res) => res.json())) as Array<{
+      events: Array<{
+        slug: string;
+      }>;
+    }>;
+
+    const slug = data?.[0]?.events?.[0]?.slug;
+
+    console.log(`Found polymarket slug for question ID: ${questionId}`, slug);
```
daywiss, line 39: can prob remove this

pages/api/get-polymarket-link.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    response.setHeader("Cache-Control", "max-age=0, s-maxage=2592000");
+    response.status(200).send({ slug });
+  } catch (e) {
+    console.error(e);
```
daywiss, line 44: might want to just add a label to this error like 'Error fetching data for poly question"

helpers/voting/projects/polymarket.ts
```diff
@@ -226,3 +227,28 @@ export function maybeMakePolymarketOptions(
...
+ */
+
+// questionId is a hash of ancillaryData + initializer EOA
+export function sanitizeAncillaryData(decodedAncillaryData: string): string {
+  const regex = /(?:,\s*)?(ooRequester|childChainId|childRequester):[^,]+/g;
+  const stripped = decodedAncillaryData.replace(regex, "");
+  return stripped;
+}
+
+export function getQuestionId({
```
daywiss, line 245: seems like a simple solution, wonder why we didnt do this before

gsteenkamp89: Couldn't query PM by questionId before, this is a new feature they implemented for us.

pages/api/get-polymarket-link.ts
gsteenkamp89: https://github.com/UMAprotocol/voter-dapp-v2/pull/356/commits/587a1a7bd68c66679cb38458b5f80e76e5b40c0f

pages/api/get-polymarket-link.ts
gsteenkamp89: https://github.com/UMAprotocol/voter-dapp-v2/pull/356/commits/587a1a7bd68c66679cb38458b5f80e76e5b40c0f

PR #357, opened 2025-03-25
fix: fix v3 assert truth title and description

daywiss:
image
image

### DIFF
--- a/web3/queries/votes/getAssertionClaim.ts
+++ b/web3/queries/votes/getAssertionClaim.ts
@@ -7,7 +7,7 @@ export async function getAssertionMadeEvents(
 ) {
   const currentBlock = await instance.provider.getBlockNumber();
   const filter = instance.filters.AssertionMade(assertionId);
-  return instance.queryFilter(filter, currentBlock - 100000);
+  return instance.queryFilter(filter, currentBlock - 1000000);
 }
 
 export async function getAssertionClaim(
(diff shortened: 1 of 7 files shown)


web3/queries/votes/getAssertionClaim.ts
```diff
@@ -7,7 +7,7 @@ export async function getAssertionMadeEvents(
 ) {
   const currentBlock = await instance.provider.getBlockNumber();
   const filter = instance.filters.AssertionMade(assertionId);
-  return instance.queryFilter(filter, currentBlock - 100000);
+  return instance.queryFilter(filter, currentBlock - 1000000);
```
mrice32, line 10: Will this not exceed block limits on some chains?

mrice32 (approved):
LGTM!

web3/queries/votes/getAssertionClaim.ts
daywiss: most likely yes, ill need to make a note to fix this later.  unfortunately on story with the current range, we cant find the claim, so had to bump it up. fortunately it still works

daywiss: made an issue for it: https://linear.app/uma/issue/UMA-2867/fix-assertion-claim-lookback-to-page-in-shorter-ranges

PR #364, opened 2025-04-11
feat(app): support new oracle bridging contracts (umip-185)

gsteenkamp89:
closes UMA-2879

## Description

Adds support for umip-185. After this upgrade the full ancillaryData will not be bridged over to L1, we need to resolve this manually.


helpers/voting/resolveAncillaryData.ts
```diff
@@ -0,0 +1,183 @@
...
+    ancillaryData: resolvedAncillaryData[i],
+  }));
+}
+
+export async function resolveAncillaryData(
+  args: Pick<RequestAddedEvent["args"], "ancillaryData" | "time" | "identifier">
+): Promise<string> {
+  const decodedAncillaryData = decodeHexString(args.ancillaryData);
+  const { ancillaryDataHash, childOracle, childChainId, childBlockNumber } =
+    extractMaybeAncillaryDataFields(decodedAncillaryData);
```
Reinis-FRP, line 78: does this guaranee the ordering of fields and that no other information is present (e.g. so no one can spoof this on L1 OOv2)

gsteenkamp89: no it doesn't I'll have to add this still

helpers/voting/resolveAncillaryData.ts
```diff
@@ -0,0 +1,183 @@
...
+    null,
+    null,
+    null,
+    null,
+    args.parentRequestId
+  );
+
+  const events = await OracleSpoke.queryFilter(
+    filter,
+    args.childBlockNumber - 1,
```
Reinis-FRP, line 170: its not strictly required to query from previous block, it can be the same as to block

gsteenkamp89: Interesting, I did try this and it fails to find anything

gsteenkamp89: @Reinis-FRP https://github.com/UMAprotocol/voter-dapp-v2/pull/364/commits/65c2211795b3bb6bbbf045c73454c0ac8bf2700e

gsteenkamp89: this should do it!

helpers/voting/resolveAncillaryData.ts
gsteenkamp89: @Reinis-FRP this should do it https://github.com/UMAprotocol/voter-dapp-v2/pull/364/commits/65c2211795b3bb6bbbf045c73454c0ac8bf2700e

helpers/voting/resolveAncillaryData.ts
Reinis-FRP: thats weird, cause they to and from should be inclusive.

Reinis-FRP (approved):

PR #365, opened 2025-04-16
fix: use l1 ancillary data for commits

daywiss:
# motivation
Commit is broken since upgrade, we need to change how commits are handled slightly

# changes
this uses the L1 ancillary data, rather than the fetched L2 original data

### DIFF
--- a/helpers/voting/formatVotes.ts
+++ b/helpers/voting/formatVotes.ts
@@ -51,7 +51,13 @@ export async function formatVotesToCommit({
       // if not, exclude this vote from the final array
       if (!selectedVote) return null;
 
-      const { identifier, decodedIdentifier, ancillaryData, time } = vote;
+      const {
+        identifier,
+        decodedIdentifier,
+        ancillaryData,
+        time,
+        ancillaryDataL1,
+      } = vote;
       // the selected option for a vote is called `price` for legacy reasons
       const price = parseVoteStringWithPrecision(
         selectedVote,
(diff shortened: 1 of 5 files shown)


helpers/voting/formatVotes.ts
```diff
@@ -90,14 +96,15 @@ export function formatVotesToReveal(decryptedVotesForUser: VoteT[]) {
...
+    const { identifier, decryptedVote, ancillaryData, ancillaryDataL1, time } =
+      vote;
     const { price, salt } = decryptedVote;
 
     return {
       identifier,
       time,
       price,
-      ancillaryData,
+      ancillaryData: ancillaryDataL1 ?? ancillaryData,
```
daywiss, line 107: this is one part im not sure about, when we hash the vote, do we use the L1 data?

web3/mutations/votes/commitVotes.ts
```diff
@@ -32,12 +39,20 @@ export async function commitVotes({ voting, formattedVotes }: CommitVotes) {
...
+      hash,
+      encryptedVote,
+    } = vote;
+
     // @ts-expect-error todo figure out why it thinks this doesn't exist
     return voting.interface.encodeFunctionData(commitVoteFunctionFragment, [
       identifier,
       time,
-      ancillaryData,
+      ancillaryDataL1 ?? ancillaryData,
```
daywiss, line 55: this is the main fix

helpers/voting/formatVotes.ts
mrice32: Yes!

gsteenkamp89 (approved):
Good catch!

daywiss:
closing in favor of #366

PR #367, opened 2025-04-16
improve: rename ancillary data for l2

daywiss:
renaming the variable from the last fix to something more reasonable, from displayAncillaryData to ancillaryDataL2


gsteenkamp89 (approved):
this is more intuitive, nice 👍

PR #378, opened 2025-06-25
sanitize titles on the frontend also

gsteenkamp89:
Follow-up of https://github.com/UMAprotocol/voter-dapp-v2/pull/377

## Changes
- Move sanitize logic to shared util folder
- Sanitize titles for display in the frontend


gsteenkamp89:
> LGTM - however, did we want to look into an "opt-in" format? I.e. allow all ascii characters 32 - 127? (the ASCII printable characters)
> 
> This way, all known printable characters can be included but we strip out these extra unicode invisible characters.

I think silently stripping out a subset of characters, while not explicitly stating this in any of our UMIPs, can create problems for anyone using the oracle. I'll keep this PR scoped to this particular issue, to play nice with Discord.

PR #398, opened 2025-09-05
Fix: allow GET method in warm-summary API handler

chrismaree:


pages/api/warm-summary.ts
```diff
@@ -202,7 +202,7 @@ export default async function handler(
   req: NextApiRequest,
   res: NextApiResponse<WarmSummaryResponse | { error: string }>
 ) {
-  if (req.method !== "POST") {
+  if (req.method !== "POST" && req.method !== "GET") {
```
gsteenkamp89, line 205: i guess the manual call to warm summaries is sending a post?

gsteenkamp89 (approved):

pages/api/warm-summary.ts
gsteenkamp89: yeah agreed!

PR #451, opened 2026-05-03
feat: serve UMIP metadata from cached API route

gsteenkamp89:
## Summary
- Adds `/api/umip-metadata` as a single server-side proxy to Contentful with per-UMIP Redis caching (1h hit TTL, 5min miss sentinel) and Vercel edge cache headers (`s-maxage=3600, stale-while-revalidate=86400`).
- Rewrites the `useContentfulData` client hook to call this endpoint instead of contacting Contentful directly. Same return shape, so `VotesContext` and `getVoteMetaData` are untouched.
- Drops `contentfulSpace`, `contentfulAccessToken`, and `contentfulEnabled` from the client-side `AppConfig` and removes the matching `NEXT_PUBLIC_CONTENTFUL_*` entries from the env schema. The API route reads `CONTENTFUL_SPACE_ID` / `CONTENTFUL_ACCESS_TOKEN`, falling back to the legacy `NEXT_PUBLIC_*` names so deploys keep working before the Vercel env vars are renamed.

## Why
N tabs × M votes worth of UMIP metadata used to mean N×M Contentful calls per minute. With this PR, all clients share one cached endpoint, collapsing that to roughly one Contentful call per UMIP per hour regardless of client count.

## Test plan
- [x] `tsc --noEmit` clean
- [x] `vitest run` — 50 tests pass
- [x] `next build` compiles successfully (page-data collection fails locally only because `.env.local` is absent in this worktree)
- [ ] Deploy to preview, hit `/api/umip-metadata?numbers=170,107` and verify shape `{ "170": {...}, "107": {...} }`
- [ ] Open active votes page on preview, confirm UMIP titles/descriptions still render
- [ ] Verify cache headers via `curl -I /api/umip-metadata?numbers=1`
- [ ] After merge: rename Vercel envs `NEXT_PUBLIC_CONTENTFUL_*` → `CONTENTFUL_*` (server-only); follow-up PR can drop the fallback

### DIFF
--- a/helpers/config.ts
+++ b/helpers/config.ts
@@ -16,8 +16,6 @@ const Env = ss.object({
   NEXT_PUBLIC_WALLET_CONNECT: ss.string(),
   NEXT_PUBLIC_GRAPH_STUDIO_API_KEY: ss.string(),
   // optional envs
-  NEXT_PUBLIC_CONTENTFUL_SPACE_ID: ss.optional(ss.string()),
-  NEXT_PUBLIC_CONTENTFUL_ACCESS_TOKEN: ss.optional(ss.string()),
   NEXT_PUBLIC_THE_GRAPH_API_KEY: ss.optional(ss.string()),
   NEXT_PUBLIC_GRAPH_ENDPOINT: ss.optional(ss.string()),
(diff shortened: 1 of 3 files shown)


gsteenkamp89:
@codex reivew

PR #453, opened 2026-05-21
fix: look up titleless discord threads under "N/A" key

gsteenkamp89:
## Summary

Titleless YES_OR_NO_QUERY requests (and any other request whose ancillary data has no `title:` token) were rendering with no Discord discussion — the lookup key never matched the bot's thread.

The dispute bot names threads `<title> - <timestamp>` using the ancillary-data title, falling back to `N/A` when none is present (e.g. `N/A - 1758134593`). The voter dapp was using the *display* title for lookup, which `getVoteMetaData` falls back to the identifier (`YES_OR_NO_QUERY`). So both the warm cron's cache write and the endpoint's cache read keyed off `YES_OR_NO_QUERY-<ts>` while the bot's thread sat under `N/A-<ts>`.

## What changed

- New helper `getDiscordThreadTitle(decodedAncillaryData)` in `helpers/voting/getVoteMetaData.ts` that returns the ancillary-data title or `"N/A"`. Exported via the `helpers` barrel.
- Frontend ([components/Panel/VotePanel/VotePanel.tsx](https://github.com/UMAprotocol/voter-dapp-v2/blob/claude/gallant-ritchie-17c013/components/Panel/VotePanel/VotePanel.tsx)) passes the helper's result into `useVoteDiscussion` instead of the display title.
- Both warm crons ([warm-discord-threads.ts](https://github.com/UMAprotocol/voter-dapp-v2/blob/claude/gallant-ritchie-17c013/pages/api/cron/warm-discord-threads.ts), [warm-summaries.ts](https://github.com/UMAprotocol/voter-dapp-v2/blob/claude/gallant-ritchie-17c013/pages/api/cron/warm-summaries.ts)) use the helper to build the cache key so the cache aligns with what the endpoint will look up.
- Defensive fallback in [pages/api/discord-thread/index.ts](https://github.com/UMAprotocol/voter-dapp-v2/blob/claude/gallant-ritchie-17c013/pages/api/discord-thread/index.ts): if an incoming `title` is empty after sanitization, substitute `"N/A"` before building the key.
- Marked `.eslintrc.json` as `"root": true` so eslint stops walking up the directory tree — the only behavioral side effect is that nested checkouts (e.g. worktrees living inside another node project) no longer trip on duplicate plugin resolution.

## Why a reviewer should care

- The `N/A` string and the timestamp format are deliberately matched to what the bot emits — changing either side requires updating both.
- `getTitleFromAncillaryData` searches for the literal `title:` token. Requests with only `q: <title>` (no `title:` keyword) intentionally fall through to `N/A`; if the bot ever changes its rules, the helper is the single place to update.
- The display title (shown on the panel) is unchanged — only the *discord lookup* title moved.

## Test plan

- [x] `yarn test` — 59/59 pass, including a new 2-case test for `getDiscordThreadTitle`.
- [x] `tsc --noEmit` clean.
- [ ] Manually verify a titleless YES_OR_NO_QUERY (e.g. the Wellington-temperature request that prompted this) now loads its Discord thread on a preview deploy.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,4 +1,5 @@
 {
+  "root": true,
   "parser": "@typescript-eslint/parser",
   "parserOptions": {
     "tsconfigRootDir": ".",
--- a/components/Panel/VotePanel/VotePanel.tsx
+++ b/components/Panel/VotePanel/VotePanel.tsx
@@ -116,6 +116,9 @@ export function VotePanel({ content }: Props) {
     : "";
 
   const voteOrigin = isOptimisticGovernorVote ? "OSnap" : origin;
+  // The discord-lookup title (ancillary-data title or "N/A") is what threads
+  // and summaries are cached under — must match what the warm-* crons write.
+  const discordLookupTitle = getDiscordThreadTitle(decodedAncillaryData);
   const {
     data: discussion,
     isFetching: discussionLoading,
@@ -123,7 +126,7 @@ export function VotePanel({ content }: Props) {
   } = useVoteDiscussion({
     identifier,
     time,
-    title: getDiscordThreadTitle(decodedAncillaryData),
+    title: discordLookupTitle,
   });
   const { data: claim } = useAssertionClaim(assertionChildChainId, assertionId);
   const { data: augmentedVoteData } = useAugmentedVoteData({
@@ -156,7 +159,10 @@ export function VotePanel({ content }: Props) {
             {
               title: "Discord Summary",
               content: (
-                <DiscussionSummary query={content} bulletins={bulletins.data} />
+                <DiscussionSummary
+                  query={{ ...content, title: discordLookupTitle }}
+                  bulletins={bulletins.data}
+                />
               ),
             },
           ]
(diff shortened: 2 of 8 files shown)


gsteenkamp89:
Addressed Codex's P2 in c6eec9a. Found one extra spot:

- **`warm-summaries.ts`**: `voteInfo.title` now uses `discordTitle` (the "N/A" fallback) instead of `metadata.title`, so the `/api/update-summary` → `/api/discord-thread` forward rebuilds the same key the cron just warmed.
- **`VotePanel.tsx` (`useDiscussionSummary`)**: same problem on the read side. The summary cache key is `discord-summary:<time>:<identifier>:<title>`, so if the cron writes under `…:N/A` and the frontend reads under `…:YES_OR_NO_QUERY`, even a generated summary won't be found. Both `useDiscussionSummary` and the inner `triggerSummaryGeneration` call now go through the same `discordLookupTitle` we already use for `useVoteDiscussion`.

Display title is unchanged.

0xjorgen (approved):

PR #455, opened 2026-05-22
Remap Polymarket bulletin owner

pumpedlunch:
## Summary
- add a hardcoded remap from deprecated Polymarket initializer 0x91430cad2d3975766499717fa0d66a78d814e5c5 to current bulletin owner 0xF43d55F3A8B7484Ed4B6931f93CB6F9eF5Dd369D
- for remapped requests, read bulletins from both the current owner and the deprecated initializer owner
- include deprecated-initializer bulletins only through May 22, 2026 at 1:00 AM PT (Unix timestamp 1779436800), while keeping current-owner bulletins unbounded
- merge and sort bulletin updates chronologically before rendering
- cover mapped, unmapped, case-insensitive, and dual-owner query behavior

## Tests
- yarn test
- yarn tsc --noEmit --pretty false
- yarn lint

### DIFF
--- a/helpers/voting/projects/__tests__/polymarket.test.ts
+++ b/helpers/voting/projects/__tests__/polymarket.test.ts
@@ -1,0 +1,50 @@
+import { describe, expect, it } from "vitest";
+
+import {
+  deprecatedPolymarketBulletinOwnerCutoffTimestamp,
+  getBulletinOwnerQueries,
+  resolveBulletinOwner,
+} from "../polymarketBulletinOwners";
+
+const deprecatedInitializer = "0x91430cad2d3975766499717fa0d66a78d814e5c5";
+const currentBulletinOwner = "0xF43d55F3A8B7484Ed4B6931f93CB6F9eF5Dd369D";
+
+describe("resolveBulletinOwner", () => {
+  it("maps the deprecated Polymarket initializer to the current bulletin owner", () => {
+    expect(resolveBulletinOwner(deprecatedInitializer)).toBe(
+      currentBulletinOwner
+    );
+  });
+
+  it("matches deprecated initializer addresses case-insensitively", () => {
+    expect(resolveBulletinOwner(deprecatedInitializer.toUpperCase())).toBe(
+      currentBulletinOwner
+    );
+  });
+
+  it("returns unmapped initializer addresses unchanged", () => {
+    const initializer = "0x1111111111111111111111111111111111111111";
+
+    expect(resolveBulletinOwner(initializer)).toBe(initializer);
+  });
+});
+
+describe("getBulletinOwnerQueries", () => {
+  it("queries the current owner and the deprecated owner with a cutoff", () => {
+    expect(getBulletinOwnerQueries(deprecatedInitializer)).toEqual([
+      { owner: currentBulletinOwner },
+      {
+        owner: deprecatedInitializer,
+        maxTimestamp: deprecatedPolymarketBulletinOwnerCutoffTimestamp,
+      },
+    ]);
+  });
+
+  it("returns one unchanged owner query for unmapped initializers", () => {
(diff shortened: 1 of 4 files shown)


mrice32:
@codex review

helpers/voting/projects/__tests__/polymarket.test.ts
```diff
@@ -0,0 +1,50 @@
+import { describe, expect, it } from "vitest";
```
droplet-rl, line 1: nit: file is named `polymarket.test.ts` but every test in it covers `polymarketBulletinOwners.ts`. There's no `polymarket.ts` test today, so this filename will be misleading if/when someone later adds tests for the real `polymarket.ts`. Consider renaming to `polymarketBulletinOwners.test.ts`.

helpers/voting/projects/polymarketBulletinOwners.ts
```diff
@@ -0,0 +1,47 @@
...
+
+const bulletinOwnerRemaps: Record<string, BulletinOwnerRemap> = {
+  "0x91430cad2d3975766499717fa0d66a78d814e5c5": {
+    owner: "0xF43d55F3A8B7484Ed4B6931f93CB6F9eF5Dd369D",
+    deprecatedOwnerUpdatesUntil:
+      deprecatedPolymarketBulletinOwnerCutoffTimestamp,
+  },
+};
+
+export function resolveBulletinOwner(initializer: string): string {
```
droplet-rl, line 24: `resolveBulletinOwner` is exported (and re-exported from `helpers/voting/projects/index.ts`) but the only non-test caller is itself. `getPolymarketBulletins` uses `getBulletinOwnerQueries`, not this. If you have a follow-up that wires `resolveBulletinOwner` into a UI surface (e.g. displaying the active owner), great — otherwise consider dropping it to keep the public surface minimal.

web3/queries/getPolymarketBulletins.ts
```diff
@@ -51,12 +74,17 @@ export async function getPolymarketBulletins(
...
-    update: utils.toUtf8String(update.update),
-  }));
+  const updates = await Promise.all(
+    ownerQueries.map(async (ownerQuery) =>
+      formatBulletinUpdates(
+        await contract.getUpdates(questionId, ownerQuery.owner),
+        ownerQuery
+      )
+    )
+  );
```
droplet-rl, line 88: Worth flagging: with the remap in place, a transient RPC failure on either `getUpdates` call now fails the whole bulletin fetch (vs. the previous single call). Probably fine — failing loud is reasonable — but if you'd rather degrade gracefully when only one owner's call fails (e.g. show what we have from the current owner if the deprecated-owner call times out), `Promise.allSettled` would do it. Not blocking.

droplet-rl (approved):
### Summary

LGTM. Small, focused, well-tested change that hardcodes the deprecated Polymarket initializer (`0x91430cad…e5c5`) → current bulletin owner (`0xF43d55F3…d369D`) remap, with bulletins from the old address bounded by a `1779436800` cutoff (verified: that's `2026-05-22 08:00 UTC` = 1 AM PDT, per the PR description).

I traced the wiring:
- `getBulletinOwnerQueries(ownerAddress)` is the only consumer, called from `web3/queries/getPolymarketBulletins.ts:77`.
- Updates from both owners are fetched in parallel via `Promise.all`, the deprecated-owner results are filtered by `maxTimestamp`, then everything is flattened and sorted chronologically. V8's sort is stable so co-timestamp ordering (current-owner first) is preserved.
- The remap key is lowercased before lookup, so initializer casing from `getInitializer()` doesn't matter. The contract call normalizes addresses, so passing the original-cased `initializer` as `owner` in the unmapped branch is fine.

A few non-blocking nits inline. Nothing here should block merge.

mrice32 (approved):
LGTM!

PR #456, opened 2026-05-22
Remap Polymarket bulletin owner

mrice32:
## Summary
Re-lands #455 onto master (was accidentally merged into `feat/update-predict-fun-tagging`).

- add a hardcoded remap from deprecated Polymarket initializer 0x91430cad2d3975766499717fa0d66a78d814e5c5 to current bulletin owner 0xF43d55F3A8B7484Ed4B6931f93CB6F9eF5Dd369D
- for remapped requests, read bulletins from both the current owner and the deprecated initializer owner
- include deprecated-initializer bulletins only through May 22, 2026 at 1:00 AM PT (Unix timestamp 1779436800), while keeping current-owner bulletins unbounded
- merge and sort bulletin updates chronologically before rendering
- cover mapped, unmapped, case-insensitive, and dual-owner query behavior

## Tests
- yarn test
- yarn tsc --noEmit --pretty false
- yarn lint

### DIFF
--- a/helpers/voting/projects/polymarketBulletinOwners.ts
+++ b/helpers/voting/projects/polymarketBulletinOwners.ts
@@ -1,7 +1,7 @@
 // Maps deprecated initializer EOAs to the current bulletin-board owner that
 // posts clarifications on their behalf. The questionId remains unchanged.
 // May 20, 2026 at 1:00 PM PT.
-export const deprecatedPolymarketBulletinOwnerCutoffTimestamp = 1779307200;
+export const deprecatedPolymarketBulletinOwnerCutoffTimestamp = 1779436800;
 
 type BulletinOwnerRemap = {
   owner: string;
--- a/helpers/voting/projects/polymarketBulletinOwners.ts
+++ b/helpers/voting/projects/polymarketBulletinOwners.ts
@@ -1,6 +1,6 @@
 // Maps deprecated initializer EOAs to the current bulletin-board owner that
 // posts clarifications on their behalf. The questionId remains unchanged.
-// May 20, 2026 at 1:00 PM PT.
+// May 22, 2026 at 1:00 AM PT.
 export const deprecatedPolymarketBulletinOwnerCutoffTimestamp = 1779436800;
 
 type BulletinOwnerRemap = {
--- a/helpers/voting/projects/polymarketBulletinOwners.ts
+++ b/helpers/voting/projects/polymarketBulletinOwners.ts
@@ -1,0 +1,47 @@
+// Maps deprecated initializer EOAs to the current bulletin-board owner that
+// posts clarifications on their behalf. The questionId remains unchanged.
+// May 20, 2026 at 1:00 PM PT.
+export const deprecatedPolymarketBulletinOwnerCutoffTimestamp = 1779307200;
+
+type BulletinOwnerRemap = {
+  owner: string;
+  deprecatedOwnerUpdatesUntil?: number;
+};
+
+export type BulletinOwnerQuery = {
+  owner: string;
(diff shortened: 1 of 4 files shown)


helpers/voting/projects/polymarketBulletinOwners.ts
```diff
@@ -0,0 +1,47 @@
...
+  "0x91430cad2d3975766499717fa0d66a78d814e5c5": {
+    owner: "0xF43d55F3A8B7484Ed4B6931f93CB6F9eF5Dd369D",
+    deprecatedOwnerUpdatesUntil:
+      deprecatedPolymarketBulletinOwnerCutoffTimestamp,
+  },
+};
+
+export function resolveBulletinOwner(initializer: string): string {
+  return bulletinOwnerRemaps[initializer.toLowerCase()]?.owner ?? initializer;
+}
```
droplet-rl, line 26: Nit (non-blocking): `resolveBulletinOwner` is exported and tested, but never imported anywhere outside the tests — `getPolymarketBulletins` uses `getBulletinOwnerQueries` instead. If there's no near-term caller, consider dropping it (and the two associated tests) so the module's surface area matches what's actually used. If you're keeping it for a planned consumer, ignore this.

helpers/voting/projects/polymarketBulletinOwners.ts
```diff
@@ -0,0 +1,47 @@
+// Maps deprecated initializer EOAs to the current bulletin-board owner that
+// posts clarifications on their behalf. The questionId remains unchanged.
+// May 22, 2026 at 1:00 AM PT.
```
droplet-rl, line 3: Nit: the `// May 22, 2026 at 1:00 AM PT.` line reads as a stray sentence between the file-level doc and the constant. Moving it directly above (or onto) the `deprecatedPolymarketBulletinOwnerCutoffTimestamp` declaration would make it clearer that it's annotating that value.

web3/queries/getPolymarketBulletins.ts
```diff
@@ -51,12 +74,17 @@ export async function getPolymarketBulletins(
...
-  }));
+  const updates = await Promise.all(
+    ownerQueries.map(async (ownerQuery) =>
+      formatBulletinUpdates(
+        await contract.getUpdates(questionId, ownerQuery.owner),
+        ownerQuery
+      )
+    )
+  );
+  return updates.flat().sort((a, b) => a.timestamp - b.timestamp);
```
droplet-rl, line 89: Worth confirming: the deprecated-owner contract call still returns *all* of that owner's updates from chain, and we filter `> maxTimestamp` client-side. That's fine today (deprecated owner shouldn't be writing much), but if the deprecated owner ever resumes posting, those post-cutoff bulletins are silently dropped on the client — no log/warning. Probably acceptable; just flagging.

droplet-rl (approved):
Looks good — clean re-land. The remap table, dual-owner fetch, cutoff filter, and chronological merge are all straightforward, and the tests cover mapped/unmapped/case-insensitive/dual-query behavior.

Verified:
- Cutoff `1779436800` = 2026-05-22 08:00 UTC = 2026-05-22 01:00 PDT, matches the description.
- Unmapped initializers return `[{ owner: initializer }]`, so existing behavior is preserved.
- `Promise.all` parallelizes the two `getUpdates` calls; `sort((a, b) => a.timestamp - b.timestamp)` is stable in modern V8, so same-timestamp updates keep current-owner-first order (current owner is queried first in `getBulletinOwnerQueries`).
- `.toLowerCase()` on lookup is consistent with the lowercased map keys.

One non-blocking nit inline about `resolveBulletinOwner`.

PR #461, opened 2026-06-15
Warn that committed votes can only be revealed by the committer

droplet-rl:
## Summary
- Adds a warning line to the delegate explanation on the **Add delegate** wallet-settings panel: committed votes can only be revealed by the wallet that committed them, so a delegator cannot reveal votes their delegate committed.
- Driven by a real user incident this round (commit by delegate, user undelegated mid-round, dapp showed "Unable to reveal" with no path out from the UI).

Only `components/pages/WalletSettings/AddDelegate.tsx` matched the exact "from" copy Lee called out. The same delegate explanation also lives in `ConnectedWallet.tsx` and `OtherWallet.tsx` (slightly different phrasing — `commit & reveal`, no `however`). Happy to mirror the warning there in a follow-up if we want consistent messaging across the settings page — let me know.

## Test plan
- [x] `pnpm dev`, open **Wallet Settings → Add delegate wallet**, confirm the warning paragraph renders below the existing copy with a blank line between them.
- [x] Confirm no layout regressions on the panel (Text is `styled.p`; `<br />` inside a `<p>` is valid).

### DIFF
--- a/components/pages/WalletSettings/AddDelegate.tsx
+++ b/components/pages/WalletSettings/AddDelegate.tsx
@@ -1,6 +1,6 @@
 import { Button } from "components";
 import { usePanelContext } from "hooks";
-import { BarText, BarWrapper, Header, Text } from "./styles";
+import { BarText, BarWrapper, Header, Text, Warning } from "./styles";
 
 export function AddDelegate() {
   const { openPanel } = usePanelContext();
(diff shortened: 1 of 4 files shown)


gsteenkamp89 (approved):
LGTM!

droplet-rl:
Thanks for the review!

PR #469, opened 2026-06-16
feat(vote-panel): opt-in auto-advance after voting (ACX-4981)

droplet-rl:
## What

Adds an opt-in **"Auto-next"** toggle to the vote panel. When enabled, selecting a quick-vote option (P1/P2/…) automatically advances to the next vote — so on mobile you no longer have to tap the next chevron after every choice.

- **Off by default**, persisted in `localStorage` (`uma-voter-auto-advance`).
- Applies on desktop + mobile (one toggle, one surface).
- Toggle lives in the existing panel `NavigationBar`, next to the "Use ←→ to navigate" hint, and only renders during the **commit phase** when a selection is possible.

Requested by @bobby. Closes [ACX-4981](https://linear.app/uma/issue/ACX-4981/auto-next-voting-on-uma-voting-on-mobile-dapp).

## How

No changes to the vote / commit / reveal flow — this just chains two behaviours that already exist (`selectVote` → `goToNextVote`) behind a preference.

- `hooks/helpers/useAutoAdvance.ts` — `localStorage`-backed preference returning `[enabled, toggle]`, modelled on `usePersistedVotes` (SSR-safe: read in an effect, not in initial state, to avoid a hydration mismatch).
- `hooks/helpers/shouldAutoAdvance.ts` — pure decision logic, unit-tested.
- `VotePanel.tsx` — a single wrapped `selectVote` shared between the click path (`VotePanelVoteInput`) and the keyboard path (`useVotePanelKeyboard`) so both advance consistently, plus the `Toggle`.
- `contexts/PanelContext.tsx` — `selectVote` gains an optional `{ skipHighlight }` (backward-compatible) so auto-advance can suppress the just-selected highlight.

## Edge cases handled

- **Deselect:** re-clicking the selected option clears it (`value === undefined`) and does **not** advance — you stay put rather than skipping an unanswered vote.
- **Last vote:** never advances when `canGoNext` is false, so you're not stranded past the end (the "Commit N votes" action lives in the list, not the panel).
- **Scroll jitter:** auto-advance suppresses the current vote's highlight so two smooth scrolls don't fire back-to-back; only the next vote highlights.
- **Chevron flash:** auto-advance calls `goToNextVote` directly (not `handleNext`), so the right-chevron's red flash stays a manual-tap affordance.
- **Keyboard parity:** number-key selection auto-advances too, since both paths share the wrapped handler.

## Testing

- `yarn test` — ✅ 70 passed (incl. 5 new for `shouldAutoAdvance`)
- `yarn lint` — ✅ no new warnings (pre-commit `eslint --max-warnings=0` passed on changed files)
- `npx tsc --noEmit` — ✅ clean; `next build` compiles successfully
- `prettier --check` — ✅ on changed files
- ⚠️ Not run in the sandbox (no env/secrets): `yarn cypress:run` (e2e) and full `next build` page-data collection, which needs `NEXT_PUBLIC_*` contract addresses. Worth a manual check on a mobile viewport before merge.

### DIFF
--- a/components/Panel/VotePanel/VotePanel.tsx
+++ b/components/Panel/VotePanel/VotePanel.tsx
@@ -1,7 +1,7 @@
 import removeMarkdown from "remove-markdown";
-import { useCallback, useRef, useState } from "react";
+import { useCallback, useMemo, useRef, useState } from "react";
 
-import { Tabs } from "components";
+import { Tabs, Toggle } from "components";
 import { config, decodeHexString, getClaimTitle } from "helpers";
 import { getOptimisticGovernorTitle } from "helpers/voting/optimisticGovernor";
 import {
@@ -11,7 +11,9 @@ import {
   useAugmentedVoteData,
   useVoteTimingContext,
   useVotePanelKeyboard,
+  useAutoAdvance,
 } from "hooks";
+import { shouldAutoAdvance } from "hooks/helpers/shouldAutoAdvance";
 import { useOptimisticGovernorData } from "hooks/queries/votes/useOptimisticGovernorData";
 import { VoteT } from "types";
 import { PanelFooter } from "../PanelFooter";
@@ -65,7 +67,7 @@ export function VotePanel({ content }: Props) {
   const canGoPrev = currentVoteIndex > 0;
   const canGoNext = currentVoteIndex < navigableVotes.length - 1;
 
-  const showVoteInput = isCommitPhase && selectVote !== undefined;
+  const [autoAdvance, toggleAutoAdvance] = useAutoAdvance();
 
   const prevButtonRef = useRef<HTMLButtonElement>(null);
   const nextButtonRef = useRef<HTMLButtonElement>(null);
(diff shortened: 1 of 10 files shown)


droplet-rl:
## Review + evaluation pass

Ran three independent review agents (logic/React-integration, vote-type coverage, adversarial) against the branch, plus added behavioural tests. Summary:

### Fixed in `27fc941`
1. **Key-repeat mass-voting (med-high).** `useVotePanelKeyboard` had no `e.repeat` guard. With auto-advance on, *holding* a number key cascaded select→advance→select across votes the user never read, silently stamping selections. Now ignores `e.repeat` on the selection path; arrow-key navigation still repeats (it only moves, never records a vote).
2. **Accessibility (med).** `Toggle` was a bare `<button>`. Now `role="switch"` + `aria-checked` + optional `aria-label` (supplied by the panel) + `type="button"`. Screen readers announce the control and its on/off state.
3. **Testability.** Extracted the select→advance chaining into a pure `makeAutoAdvanceSelectVote()` factory with behavioural tests (call ordering, `skipHighlight` flag, no-advance on deselect/last-vote/disabled, undefined passthrough).

### Confirmed sound (no change needed)
Fresh-index advance (no stale closure), single-render batching, keyboard/click parity, reveal-phase gating, hydration-safe persistence, localStorage robustness (garbage/`"TRUE"`/quota/private-mode all default to off), off-by-one boundaries, and coverage across custom-value, governance, OSnap, Polymarket, single-vote, and delegate/delegator cases.

### By design (flagging, not fixing)
- The "Early request" magic-number option auto-advances like P1/P2 — it's a real selectable choice, so this is consistent. The true "Custom" entry is excluded and never auto-advances.
- The "Auto-next" toggle shows in commit phase whenever the panel has navigation, even if the *current* vote has no quick options — it's a session preference that applies as you navigate.

### Known limitation (follow-up candidate)
- **Focus management:** after an auto-advance the just-clicked P-button unmounts, so keyboard/AT focus drops to `<body>`. Low impact for the primary mobile/touch use case; a proper fix (move focus to the next vote's option) is worth a follow-up but adds complexity disproportionate to this papercut.

### Checks (local)
`yarn test` ✅ 76 passed (11 new across the two helpers) · `tsc --noEmit` ✅ · `yarn lint` ✅ (pre-commit `eslint --max-warnings=0` passed) · `prettier --check` ✅. Cypress e2e + full `next build` page-data still need `NEXT_PUBLIC_*` secrets not present in the sandbox.

droplet-rl:
Closing per @0xjorgen's call on the Linear issue ([ACX-4981](https://linear.app/uma/issue/ACX-4981/auto-next-voting-on-uma-voting-on-mobile-dapp)).

The mobile two-tap papercut is real, but auto-advance isn't necessarily the right fix — the priority is giving clear UI feedback that the selected vote matches the user's intent. The quick-voting flow is being handed to the upcoming redesign for a holistic solution.

Leaving the branch (`droplet/acx-4981-auto-advance-voting`) intact so this implementation is recoverable as a reference for that work.

PR #472, opened 2026-07-08
fix: return early in handleApiError after sending HttpError response

gsteenkamp89:
## What changed

`handleApiError()` in `pages/api/_utils/errors.ts` was missing a `return` after the `instanceof HttpError` branch, so every handled HTTP error fell through to the fallback path, which:

- logged the error a second time as "Unexpected API error" (mislabeling handled 4xx/5xx errors as unexpected 500s in the logs), and
- called `response.status(500).json(...)` after the response had already been sent, producing `ERR_HTTP_HEADERS_SENT` warnings in serverless logs.

This PR adds an early `return` after the `HttpError` response is sent, plus a minimal unit test (`pages/api/_utils/__tests__/errors.test.ts`) covering both paths:

- an `HttpError` sends exactly one response with its status code (fails on the old code — `status` was called twice)
- a plain `Error` still takes the 500 fallback path

## Notes for reviewers

No behavior change for clients — the first response was always the one that reached them. This only fixes the duplicate send attempt and the misleading log entries.

### DIFF
--- a/pages/api/_utils/__tests__/errors.test.ts
+++ b/pages/api/_utils/__tests__/errors.test.ts
@@ -1,0 +1,41 @@
+import { NextApiResponse } from "next";
+import { describe, expect, it, vi } from "vitest";
+import { handleApiError, HttpError } from "../errors";
+
+function makeMockResponse() {
+  const json = vi.fn();
+  const status = vi.fn().mockReturnValue({ json });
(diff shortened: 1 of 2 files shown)


droplet-rl:
Superseded by #474 — identical commit, recreated with Droplet as author per @gsteenkamp89's request in Slack.

PR #476, opened 2026-07-15
fix: resolve L2 ancillary data on demand instead of for all past votes

gsteenkamp89:
## Problem

Opening the voter dapp spammed `/api/resolve-l2-ancillary-data` in what looked like an endless loop. Measured on prod: **~3,900 requests on load** (one per past vote in UMA's history), repeated **every 60 seconds** via `usePastVotes`' `refetchInterval`, and again on every window focus. Most repeats were absorbed by the browser cache, but every cold visitor slams the endpoint with the full sweep, and the dapp burns main-thread time issuing thousands of fetches per minute.

Root cause: `getPastVotesV2Lightweight` called `resolveAncillaryDataForRequests` over the entire resolved-request history, and the client helper called the API unconditionally — even for the ~1,000 requests that aren't bridged and need no resolution at all.

## Fix

Resolution is now on demand, scoped to what's actually rendered:

- **`getPastVotesV2Lightweight`** no longer resolves anything; the list ships raw ancillary data (`ancillaryDataL2` defaults to the raw value).
- **New `useResolvedVoteList` hook** resolves bridged ancillary data for only the votes passed to it, via one react-query entry per vote with `staleTime: Infinity` (resolution is immutable). Applied at the three render sites: the paginated past-votes page, the homepage recent-5, and the vote panel (which shows its existing loading state while resolving, so deeplinks and history clicks still get fully resolved data).
- **`resolveAncillaryData` (client helper)** short-circuits without any HTTP call when the ancillary data has no bridged (L2) fields, and memoizes in-flight/settled resolutions per session. This also shrinks the remaining query-level resolution done for active/upcoming votes.
- **`getVoteMetaData`** returns identifier-derived metadata for unresolved bridged references instead of trying to parse them (previously this logged `JSON.parse` errors for every unresolved `MULTIPLE_VALUES` vote on every render).

`uniqueKey` is derived from the *raw* ancillary data, so deeplinks, vote lookups, and contentful/admin matching are unaffected.

## Result

Verified against a local dev server: page loads now make **~50 resolution calls** (active votes + visible rows) instead of ~3,900, titles and option labels for bridged Polymarket votes render correctly, and the per-minute refetch cycle no longer re-hits the endpoint at all (each vote resolves at most once per session).

## Notes for review

- `HistoryPanel` rows don't render titles, so they intentionally don't resolve; clicking through to a vote resolves it in the panel.
- Active/upcoming vote queries still resolve at query level — those sets are small (~40) and now benefit from the non-bridged skip + memoization.
- Typecheck, lint, and the vitest suite (102 tests) pass.

### DIFF
--- a/components/Panel/VotePanel/VotePanelWithLazyLoad.tsx
+++ b/components/Panel/VotePanel/VotePanelWithLazyLoad.tsx
@@ -1,6 +1,7 @@
 import { LoadingSpinner } from "components";
 import {
   usePastVoteDetails,
+  useResolvedVoteList,
   useUserVotingAndStakingDetails,
   useAccountDetails,
   useDelegationContext,
@@ -14,7 +15,12 @@ interface Props {
   content: VoteT;
 }
 
-export function VotePanelWithLazyLoad({ content }: Props) {
+export function VotePanelWithLazyLoad({ content: rawContent }: Props) {
+  // votes from the past votes list carry unresolved (bridged) ancillary data
+  const {
+    resolvedVotes: [content],
+    isResolving,
+  } = useResolvedVoteList(useMemo(() => [rawContent], [rawContent]));
   const { address: userAddress } = useAccountDetails();
   const { isDelegate, delegatorAddress } = useDelegationContext();
   const userOrDelegatorAddress = isDelegate ? delegatorAddress : userAddress;
@@ -59,7 +65,7 @@ export function VotePanelWithLazyLoad({ content }: Props) {
     return content;
   }, [content, detailedVote, needsDetailedData]);
 
-  if (needsDetailedData && isLoading) {
+  if ((needsDetailedData && isLoading) || isResolving) {
     return (
       <LoadingWrapper>
         <LoadingSpinner size={40} variant="black" />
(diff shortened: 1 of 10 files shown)


gsteenkamp89:
Verified end-to-end on the Vercel preview deployment (production build):

| Scenario | Baseline (prod today) | This branch (preview) |
|---|---|---|
| Opening /past-votes | ~3,900 calls to `/api/resolve-l2-ancillary-data` | 66 calls (active votes + first page) |
| Paginating to page 2 (20 rows) | 0 new (already fetched all upfront) | +19 (only the bridged rows on that page) |
| Steady state (per-minute `usePastVotes` refetch) | ~3,855 calls **every minute**, forever | **0** — no endpoint calls after 4+ minutes on page; subgraph refetches continue but resolution is cached per vote |

Bridged Polymarket rows render correct titles and correct-vote labels (e.g. "Over"/"Odd") on the list and in the panel.

gsteenkamp89:
Superseded by #478, which includes everything here — the lazy on-demand resolution (as `useVotesWithOnScreenData`, also wired through the #475 deeplink flow, which this branch predates), the `getVoteMetaData` unresolved-stamp handling, and as of af5e7b7 the client-side short-circuit for non-bridged requests. Closing in favor of #478.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #480, opened 2026-07-24
fix: abort hung subgraph requests so retries and RPC fallbacks can fire

gsteenkamp89:
## Motivation

During the 2026-07-24 Goldsky outage, subgraph requests **hung** instead of erroring. Nothing above the fetch layer copes with a promise that never settles:

- react-query only retries on **rejection** — a pending promise keeps queries in `isLoading` forever, with no error state and no console warning
- the upcoming-votes RPC fallback lives in a `catch` block, so the one fallback protecting a live-vote path was structurally unreachable during exactly the outage type it was built for
- API routes (`resolve-deeplink`, `augment-request-gql`) hung until Vercel killed the function (504, full execution window billed)

## Changes

- **`helpers/util/subgraphRequest.ts`** (new): wraps `graphql-request` with an `AbortController` timeout. On timeout the fetch aborts and the promise **rejects** with a descriptive error, which re-arms everything that already exists: react-query retries with backoff, the upcoming-votes catch falls back to the RPC scan, `warnOnce` logs, error states render. Real errors pass through unchanged.
- **`constant/graph/requestTimeouts.ts`** (new): 20s for client fetches, 10s for API routes. Deliberately generous (~5–10× the slowest healthy response, the 1000-row past-votes pages) — the goal is only to convert "hangs forever" into a rejection, never to cut off a slow success.
- Every subgraph call site now goes through the wrapper: `fetchAllDocuments` (per page), past votes list/details/user votes, upcoming votes, active vote results, globals (APR), user voting & staking details, OptimisticGovernor proposal data, and the `resolve-deeplink` + `augment-request-gql` routes (server timeout).

Deliberately untouched: RPC/ethers calls and wallet interactions — different providers and failure modes, and the critical commit/reveal path is already subgraph-independent.

## Testing

- New unit test covers the three wrapper behaviors: passthrough success, rejection on hang, real errors not masked as timeouts.
- `tsc --noEmit` clean; full vitest suite 105/105.

### DIFF
--- a/constant/graph/requestTimeouts.ts
+++ b/constant/graph/requestTimeouts.ts
@@ -1,0 +1,11 @@
+// Subgraph outages can hang connections open instead of erroring. Nothing
+// above the fetch layer copes with a promise that never settles — react-query
+// only retries on rejection and RPC fallbacks live in catch blocks — so
+// subgraph requests are aborted after these windows. Values are deliberately
+// generous (~5-10x the slowest healthy response) so a slow success is never
+// cut off; the goal is only to turn "hangs forever" into a rejection.
+export const SUBGRAPH_REQUEST_TIMEOUT_MS = 20_000;
+
+// tighter on API routes so a hung subgraph returns a controlled error
+// instead of burning the function's whole execution window
+export const SERVER_SUBGRAPH_REQUEST_TIMEOUT_MS = 10_000;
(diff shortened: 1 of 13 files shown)


gsteenkamp89:
Reopened and brought up to date with `master` (merged in #482–#485, #487; commit `fbc81480`). Clean merge, no conflicts, and no new subgraph call sites bypass the timeout wrapper. `tsc --noEmit` clean; full suite 127/127.

The problem this fixes is still unaddressed on `master`: all subgraph fetches use plain `graphql-request` with no timeout, and the upcoming-votes RPC fallback still lives in a `catch` block a hang can never reach.

